PR #2, opened 2017-10-04
disty 0.1.0 (new formula)

sjackman:
Disty McMatrixface
Compute a distance matrix from a core genome alignment file.


sjackman:
```
==> brew audit linuxbrew/bioinformatics/disty --online --new-formula
  * Dependency 'zlib' may be unnecessary as it is provided by macOS; try to build this formula without it.
  * GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)
```
https://circleci.com/gh/Linuxbrew/homebrew-bioinformatics/3
Otherwise successful.

PR #41, opened 2018-02-19
fwdpp: import from brewsci/science

sjackman:


sjackman:
```
g++-5: internal compiler error: Killed (program cc1plus)
```
Exceeded available memory on CircleCI.

PR #48, opened 2018-02-21
ascp 3.7.7

tseemann:
This is a CLI for `ascp` from Aspera.  Yes, it's a binary installer. It works on Mac and Linux.

-----

### DIFF
--- a/Formula/ascp.rb
+++ b/Formula/ascp.rb
@@ -1,0 +1,54 @@
+class Ascp < Formula
+  desc "Aspera command-line download client"
+  homepage "https://example.com/"
+  if OS.mac?
+    url "https://download.asperasoft.com/download/sw/cli/3.7.7/aspera-cli-3.7.7.608.927cce8-mac-10.7-64-release.sh"
+    sha256 "c6f7af506a4de8858b8b40e63883e671926af7b43160d4fb6765790c56b299ba"
+  else
+    url "https://download.asperasoft.com/download/sw/cli/3.7.7/aspera-cli-3.7.7.608.927cce8-linux-64-release.sh"
+    sha256 "83efd03b699bdb1cac6c62befb3812342d6122217f4944f732ae7a135d578966"
+  end
+  version "3.7.7"
+
+  unless OS.mac?
+    depends_on "patchelf" => :build
+  end
+
+  def install
+    # Deduce download name from URL
+    installer = stable.url.sub %r{^.*/}, ""
+    # installer = "ascp-#{version}.sh"
+    # Patch in preferred install location (can't specify on cmdline)
+    idir = OS.mac? ? '"$HOME/Applications"' : "~/.aspera"
+    inreplace installer, "INSTALL_DIR=#{idir}", "INSTALL_DIR=#{prefix}"
+    system "sh", installer
+    # Move everything up a folder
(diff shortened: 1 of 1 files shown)


tseemann:
Here we go again with the classic CATCH-22

`home "https://` => 
image
(as certificate expired)

`home => "http://"
**must use https**

Help @sjackman ?

sjackman:
On both CircleCI and TravisCI
```
The URL https://asperasoft.com/ is not reachable
```
https://travis-ci.org/brewsci/homebrew-bio/builds/344170727#L2132

sjackman:
Intersting. https://asperasoft.com gives a 301 redirect to http://asperasoft.com

```
❯❯❯ curl -D /dev/stdout https://asperasoft.com         master
HTTP/1.1 301 Moved Permanently
Server: nginx
Date: Wed, 21 Feb 2018 05:37:03 GMT
Content-Type: text/html
Content-Length: 178
Connection: keep-alive
Location: http://asperasoft.com/
X-UA-Compatible: IE=edge,chrome=1

<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

sjackman:
On Linux…
```
curl https://asperasoft.com
curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```

sjackman:
I'd say this is a false positive in `brew audit`, though pretty unusual.

ebzheng:
Parachuting in as an observer - the binary installer for me refused to run unless I was running as root. If you actually edit it in an editor, you can see the check for root, as it's a couple of bash lines.

A second issue - for me at least, the binary Linux installer didn't give me the necessary default OpenSSL keys needed to connect to NCBI services.

I have workarounds - though I don't know if they're kosher: for the binary installer, I ran it using sudo on a system where I did have sudo access, and then simply copied the unpacked binary to the target system. To get the keys, I installed ascp on a windows system and copied them.

Perhaps this could be used to make a bottle, but I don't know if that would run afoul of licensing?

tseemann:
@ebzheng 
1. this all works as non-root for me; it was in my personal brew repo for 2 years.
2. the keys are in the `etc` folder of the installation, i just give the full path to it with `-i` on `ascp`

sjackman:
> the keys are in the etc folder of the installation, i just give the full path to it with -i on ascp

If that's a common issue, you could put it in `caveats`.

ebzheng:
On looking over the source, I see that now! I guess I was a bit trigger-happy (admittedly I spent a while last week trying to figure out the install - this would have been extremely helpful had I known about it!)

tseemann:
I did put it in `caveats` :)

PR #67, opened 2018-02-26
sga: import from brewsci/science

sjackman:


sjackman:
```
VariantIndex.cpp:89:16: error: call to 'abs' is ambiguous
```
https://gist.github.com/sjackman/2ec0bd6489b19ecdfee781938ec6e79b#file-03-make-L52

PR #89, opened 2018-03-02
kent-tools: import from brewsci/science, update to kent-tools 361

lukauskas:

(I hope)

-----

This moves `kent-tools` from the `homebrew-science` repository. 
At the same time it bumps the version number to 361.

See https://github.com/brewsci/homebrew-science/pull/64

### DIFF
--- a/Formula/kent-tools.rb
+++ b/Formula/kent-tools.rb
@@ -16,6 +16,7 @@ class KentTools < Formula
   depends_on "libpng"
   depends_on "mysql"
   depends_on "openssl"
+  depends_on "rsync" unless OS.mac?
   depends_on "util-linux" unless OS.mac?
   depends_on "zlib" unless OS.mac?
 
--- a/Formula/kent-tools.rb
+++ b/Formula/kent-tools.rb
@@ -4,21 +4,15 @@ class KentTools < Formula
   url "http://hgdownload.soe.ucsc.edu/admin/exe/userApps.v361.src.tgz"
   sha256 "4eef556e9a6191b2f7f76b1b3d01bea356ae5648e9d00e79ddc8e2a61cd37e85"
   head "git://genome-source.cse.ucsc.edu/kent.git"
-  # tag "bioinformatics"
-
-  bottle do
-    cellar :any
-    sha256 "04b38d218eb87829f33baae42d453b872aa2de986a330754445b2fc12d60fb93" => :high_sierra
-    sha256 "2b2faf1a428b99541c20dc4136c8ea6d6f4832607364478e537c278baeb4f676" => :sierra
-    sha256 "6bf095413b6aa2de56551c51a81273ced85efae0b59905492fccdbc08e7d2b8c" => :el_capitan
-  end
 
   depends_on "libpng"
   depends_on "mysql"
   depends_on "openssl"
-  depends_on "rsync" unless OS.mac?
-  depends_on "util-linux" unless OS.mac?
-  depends_on "zlib" unless OS.mac?
+  unless OS.mac?
+    depends_on "rsync"
+    depends_on "util-linux"
+    depends_on "zlib"
+  end
 
   def install
     libpng = Formula["libpng"]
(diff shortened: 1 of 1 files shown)


lukauskas:
Any ideas on why travis-ci is failing?

sjackman:
The TravisCI failure is…
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```
I'll ignore this failure.
https://travis-ci.org/brewsci/homebrew-bio/builds/348280853
https://api.travis-ci.org/v3/job/348280855/log.txt

sjackman:
Thanks for your contribution to Brewsci/bio, Saulius! I see that it's your first, and I hope that there's more to come! 🏅

sjackman:
I've created a handy script to somewhat automate these migrations. Try it out next time!
```sh
brew tap linuxbrew/developer
brew migrate-formula --tap=brewsci/bio brewsci/science/kent-tools
```

Formula/kent-tools.rb
```diff
@@ -0,0 +1,74 @@
+class KentTools < Formula
+  desc "Utilities for the UCSC Genome Browser"
+  homepage "https://genome.ucsc.edu/"
+  url "http://hgdownload.soe.ucsc.edu/admin/exe/userApps.v361.src.tgz"
+  sha256 "4eef556e9a6191b2f7f76b1b3d01bea356ae5648e9d00e79ddc8e2a61cd37e85"
+  head "git://genome-source.cse.ucsc.edu/kent.git"
+  # tag "bioinformatics"
```
sjackman: You can delete this line, since all formulae in this repo are bioinformatics.

Formula/kent-tools.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  bottle do
+    cellar :any
+    sha256 "04b38d218eb87829f33baae42d453b872aa2de986a330754445b2fc12d60fb93" => :high_sierra
+    sha256 "2b2faf1a428b99541c20dc4136c8ea6d6f4832607364478e537c278baeb4f676" => :sierra
+    sha256 "6bf095413b6aa2de56551c51a81273ced85efae0b59905492fccdbc08e7d2b8c" => :el_capitan
+  end
```
sjackman: Please delete the bottle block before submitting to Brewsci/bio.

Formula/kent-tools.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  depends_on "libpng"
+  depends_on "mysql"
+  depends_on "openssl"
+  depends_on "rsync" unless OS.mac?
+  depends_on "util-linux" unless OS.mac?
+  depends_on "zlib" unless OS.mac?
```
sjackman: A single blocks is preferred, like so…
```ruby
unless OS.mac?
  depends_on "rsync"
  depends_on "util-linux"
  depends_on "zlib"
end
```

Formula/kent-tools.rb
```diff
@@ -0,0 +1,74 @@
...
+    args << "PNGLIB=-L#{libpng.opt_lib} -lpng -lz"
+    args << "PNGINCL=-I#{libpng.opt_include}"
+
+    # On Linux, depends_on :mysql looks at system MySQL so check if Homebrew
+    # MySQL already exists. If it does, then link against that. Otherwise, let
+    # kent-tools link against system MySQL (see kent/src/inc/common.mk)
+    if mysql.installed?
+      args << "MYSQLINC=#{mysql.opt_include}/mysql"
+      args << "MYSQLLIBS=-lmysqlclient -lz"
+    end
```
sjackman: Remove the comment and conditional here. Always link against the brewed `mysql`.

lukauskas:
Thanks for the tips @sjackman!

I think I've touched upon most of them at this push.
While we're fixing the formula, there's one thing that I do not like in it, these lines that remove binaries that conflict with formula `blat`.

```
cd bin do
      blat_bin = %w[blat faToTwoBit gfClient gfServer nibFrag pslPretty
                    pslReps pslSort twoBitInfo twoBitToFa]
      rm blat_bin
      mv "calc", "kent-tools-calc"
    end
```

Now I am not sure why this is done the way it is done right now.
But maybe there is a way to create an option that would prevent formula from removing the `blat_bin` binaries, and then make this formula conflict with `blat` formula if this option is selected?

I'm afraid I do not know enough homebrew basics to attempt that.

sjackman:
For future PRs, please use `git commit --amend` and `git push --force` to update your commit, rather than add more commits to the same PR.

sjackman:
Merged. Thanks for your first contribution to Brewsci/bio! I hope that there are more coming!

sjackman:
> Now I am not sure why this is done the way it is done right now.
But maybe there is a way to create an option that would prevent formula from removing the blat_bin binaries, and then make this formula conflict with blat formula if this option is selected?

It would be redundant to have both `kent-tools` and `blat` provide `blat`, and they would conflict. You can install both `kent-tools` and `blat` if you want both. Any issue with that?

lukauskas:
Thanks for merging this!

> It would be redundant to have both kent-tools and blat provide blat, and they would conflict. You can install both kent-tools and blat if you want both. Any issue with that?

The only problem is that `blat` hasn't been updated since November 2014. See the timestamps [here](http://hgwdev.cse.ucsc.edu/~kent/src/).
While kent tools, which include `blat` binaries (as well as helper binaries included in both like `twoBitToFa`) has been updated regularly since then.
They are not very clear in their changelogs so it is hard to tell if any `blat` binaries in `kent-tools` were updated since 2014 or whether it was some other binaries that were updated.

I'm a bit worried that maybe the `blat` package has been deprecated by them with the idea to bundling it with kent tools instead, but cannot really tell at the moment.

sjackman:
I'd suggest extracting the two source tarballs and diffing them with `diff -r`.

PR #114, opened 2018-03-13
e-mem: import from brewsci/science

emreerhan:


Formula/e-mem.rb
```diff
@@ -0,0 +1,22 @@
+class EMem < Formula
+  # cite Khiste_2015 "https://doi.org/10.1093/bioinformatics/btu687"
```
sjackman: There ought to be a colon here after `Khiste_2015`.

sjackman:
Merged in dccf4adad86c102161df19eccee047a795d75adc. Thanks, Emre!

sjackman:
Merged in dccf4adad86c102161df19eccee047a795d75adc. Thanks again, Emre!

PR #127, opened 2018-03-16
CircleCI: Update homebrew/core

sjackman:


.circleci/config.yml
```diff
@@ -12,6 +12,9 @@ jobs:
           cd /home/linuxbrew/.linuxbrew/Homebrew
           git fetch origin --unshallow --tags
           git reset --hard origin/master
+          cd /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core
```
maxim-belkin: you could use `cd $(brew --repo linuxbrew/core)` which is a bit shorter

sjackman: True. When something is never expected to change (hopefully), I like the clarity of specifying the precise value. I don't mind either, really.

PR #148, opened 2018-03-21
kraken: import from brewsci/science

emreerhan:


tseemann:
@emreerhan thank you for the contribution

Would you be able to upgrade it to kraken 1.1 please?
https://github.com/DerrickWood/kraken/releases

Jennifer has taken over from Derrick and github is the main home now.

emreerhan:
Thanks @tseemann . My spaced seeds work isn't quite at a presentable stage yet. I'll be wrapping up that project soon, so I'll polish up the github page with all my results.

tseemann:
@emreerhan thanks for making the changes
One thing I tweaked was to use a "positive test" rather than a "negative" one. 

Normally I would do something like this:
```
assert_match version.to_s, shell_output("#{bin}/kraken --version 2>&1")
```
which should return 0, and match the version we expected to install.

In this case, they forgot to bump the version :-)

I also filed the wrong vers
https://github.com/DerrickWood/kraken/issues/117

So I changed it to run `--help` which returns 0.

PR #149, opened 2018-03-21
bowtie: import from brewsci/science

emreerhan:


tseemann:
I think this formula needs to be modernized and fixed before being accepted.

The main problem is that the test sets `PERL5LIB` at run time, which means the scripts will not work for people after installation as it.

@sjackman and I are experimenting with using `cpanm` to insttall Perl modules in a "venv" per package.

Have a look at `circos` for an example:
https://github.com/brewsci/homebrew-bio/blob/master/Formula/circos.rb

emreerhan:
Hey @tseemann !
You'll have to forgive me, I'm still learning the conventions for these scripts. It seems like bowtie does not actually use any perl. The perl dependencies were there for their perl benchmarking scripts in the share directory. @sjackman recommended I remove the perl dependencies completely and grep for "usage" in their help commands for the test.

Formula/bowtie.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "make", "install", "prefix=#{prefix}"
+
+    doc.install "MANUAL", "NEWS", "TUTORIAL"
+    pkgshare.install "scripts", "genomes", "indexes", "reads"
+
+    inreplace pkgshare/"scripts/test/simple_tests.pl" do |s|
+      s.gsub! "$bowtie = \"\"", "$bowtie = \"#{bin}/bowtie\""
+      s.gsub! "$bowtie_build = \"\"", "$bowtie_build = \"#{bin}/bowtie-build\""
+      s.gsub! "--debug", ""
+    end
```
sjackman: Since we're no longer running these tests, you can remove this patch.

sjackman:
Looks good, Emre. Please squash the two commits down to a single commit. You can use `git commit --amend` for fixes like this to avoid having to squash later.

tseemann:
1. I would suggest testing `--help` and leaving return code empty (`0` = good).

2. Maybe use a loop
```
Dir["#{bin}/*"].each do |exe|
  assert_match "usage:", shell_output("#{exe} --help 2>&1")
end
```

```
test do
+    assert_match "usage", shell_output("#{bin}/bowtie 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-align-l 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-align-s 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-build 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-build-l 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-build-s 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-inspect 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-inspect-l 2>&1", 1)
+    assert_match "usage", shell_output("#{bin}/bowtie-inspect-s 2>&1", 1)
+  end
```

sjackman:
Merged. Thanks again, Emre!

PR #167, opened 2018-03-25
kmc: Does not depend on gcc and fix bottle block

sjackman:
No linkage to `gcc` and so no run time dependency.
```
❯❯❯ brew linkage kmc
System libraries:
  /usr/lib/libSystem.B.dylib
  /usr/lib/libbz2.1.0.dylib
  /usr/lib/libz.1.dylib
```


sjackman:
Merged without updating the bottle.

PR #200, opened 2018-04-06
jellyfish 2.2.9 (migrate + update)

tseemann:

-----

### DIFF
--- a/Formula/jellyfish.rb
+++ b/Formula/jellyfish.rb
@@ -1,0 +1,24 @@
+class Jellyfish < Formula
+  # cite Marcais_2001: "https://doi.org/10.1093/bioinformatics/btr011"
+  desc "Fast, memory-efficient counting of DNA k-mers"
+  homepage "http://www.genome.umd.edu/jellyfish.html"
+  url "https://github.com/gmarcais/Jellyfish/releases/download/v2.2.9/jellyfish-2.2.9.tar.gz"
+  sha256 "854511d06f04cc2a8459d3f171862381d98430494194ff6f1970d1ffb74c80a8"
+
+  depends_on "pkg-config" => :build
+
+  depends_on "htslib"
+  depends_on "samtools"
+  
+  depends_on :linux
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    assert_match version.to_s, shell_output("#{bin}/jellyfish --version")
+  end
+end


tseemann:
This was meant to be fixed by using `gcc` but alas no:
```
==> make
/Applications/Xcode-9.app/Contents/Developer/usr/bin/make  all-am
  CXX      lib/rectangular_binary_matrix.lo
  CXX      lib/mer_dna.lo
  CXX      lib/storage.lo
  CXX      lib/allocators_mmap.lo
lib/rectangular_binary_matrix.cc: In static member function 'static uint64_t* jellyfish::RectangularBinaryMatrix::alloc(unsigned int, unsigned int)':
lib/rectangular_binary_matrix.cc:35:14: error: 'aligned_alloc' was not declared in this scope
   if(!(mem = aligned_alloc(sizeof(uint64_t) * 2, alloc_columns * sizeof(uint64_t))))
              ^~~~~~~~~~~~~
  CXX      lib/misc.lo
make[1]: *** [lib/rectangular_binary_matrix.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [all] Error 2
```

sjackman:
> `void *aligned_alloc( size_t alignment, size_t size );` (since C11)

http://en.cppreference.com/w/c/memory/aligned_alloc

I haven't been able to get `aligned_alloc` to work on macOS 10.11 El Capitan. I'll try it on macOS 10.13 High Sierra at home later and get back to you.

Formula/jellyfish.rb
```diff
@@ -0,0 +1,24 @@
+class Jellyfish < Formula
+  # cite Marçais_2001: "https://doi.org/10.1093/bioinformatics/btr011"
+  desc "Fast, memory-efficient counting of DNA k-mers"
+  homepage "http://www.genome.umd.edu/jellyfish.html"
+  url "https://github.com/gmarcais/Jellyfish/releases/download/v2.2.9/jellyfish-2.2.9.tar.gz"
+  sha256 "854511d06f04cc2a8459d3f171862381d98430494194ff6f1970d1ffb74c80a8"
+
+  depends_on "gcc" if OS.mac? # Fix error: use of undeclared identifier 'aligned_alloc'
+  depends_on "zlib" unless OS.mac?
```
sjackman:
```ruby
if OS.mac?
  # Fix error: use of undeclared identifier
  depends_on "gcc"
else
  depends_on "zlib"
end
```

Formula/jellyfish.rb
```diff
@@ -0,0 +1,24 @@
...
+    
+  fails_with :clang # Fix error: use of undeclared identifier 'aligned_alloc'
+
+  needs :cxx11
```
sjackman: I usually put `fails_with` and `needs` above `depends_on`. Does `brew audit` have a preference?

Formula/jellyfish.rb
```diff
@@ -0,0 +1,24 @@
+class Jellyfish < Formula
+  # cite Marçais_2001: "https://doi.org/10.1093/bioinformatics/btr011"
```
sjackman: I'd avoid UTF-8 characters in formulae. Better to stick to ASCII.

Formula/jellyfish.rb
tseemann: Brew audit doesn't care. I've been putting `needs` before `install` as that is where it seemed to be in previous formula.

sjackman: I'm fine with either.

Formula/jellyfish.rb
tseemann: Turns out it doesn't need `zlib`.

Formula/jellyfish.rb
tseemann: Fixed.

tseemann:
Could I try appending `-stdc11` to `CFLAGS` ?  
Any chance that would work?
What's the normal way to add stuff like that in `./configure` ?

tseemann:
I tried
```
   ENV.append "CFLAGS", "-std=c11"
    system "./configure", "--prefix=#{prefix}"
```
No bueno!

sjackman:
As far as I can tell `aligned_alloc` does not work on macOS. I'd suggest reporting the issue upstream to the Jellyfish developers and for now `depends_on :linux`
`gcc` on macOS still uses the Apple-provided `libc`, which does not implement  `aligned_alloc`, so `gcc` won't work either on macOS.

PR #211, opened 2018-04-11
nxtrim: import from brewsci/science

sjackman:


sjackman:
```
==> brew linkage --test brewsci/bio/nxtrim
Missing libraries:
  libz.so.1
No unwanted system libraries
==> FAILED
```
https://travis-ci.org/brewsci/homebrew-bio/builds/365262711#L2402
Depends on `zlib`

PR #231, opened 2018-04-16
mrbayes: import from brewsci/science

jonchang:


sjackman:
```
Error: No available formula with the name "beagle" (dependency of brewsci/bio/mrbayes)
```
You'll need to first migrate `beagle`
https://travis-ci.org/brewsci/homebrew-bio/builds/367280613#L1600

jonchang:
Enabled `beagle` by default since `beagle` no longer requires fiddling with CUDA and openCL seems to work "ok".

PR #233, opened 2018-04-18
beagle: import from brewsci/science

jonchang:


jonchang:
Requires `libtool` at runtime for some reason.

sjackman:
That's unusual. What was the error message?

jonchang:
`brew linkage` failed after uninstalling build time dependencies on Linux, not finding `libltdl`. I tried `ENV.refurbish_args` but that didn't work.

jonchang:
Looks like `beagle` uses `libtool` to dynamically load shared libraries at runtime... `libhmsbeagle/plugin/LibtoolSharedLibrary.h`: `This header implements unix libraries using libtool loading instead of the standard dlopen() interface`

jonchang:
BEAST and MrBayes both seem happy with `beagle` without run-time libtool. Disabling it.

PR #260, opened 2018-05-01
augustus: import from homebrew/core

suujia:


sjackman:
We won't migrate `augustus` to Brewsci/bio, since it's already in Homebrew/core, which is fine; it can stay there. We do need to bump the revision for Linux though in Linuxbrew/core. Note that Linuxbrew/core is a fork of Homebrew/core, which adds support for Linux.

PR #261, opened 2018-05-01
gaemr: import from brewsci/science

suujia:


sjackman:
```
==> brew test brewsci/bio/gaemr --verbose
…
ImportError: No module named Bio
```
https://travis-ci.org/brewsci/homebrew-bio/builds/373707260#L1637
`gaemr` depends on Biopython. Brewsci/bio doesn't yet have a package for Biopython, but we may want to create one. I'll look into it and get back to you.

sjackman:
We won't create a formula for Biopython. It's only a few megs. We can vendor it in when needed.

suujia:
On linux: 
```
==> PYTHONPATH=/home/linuxbrew/.linuxbrew/Cellar/gaemr/1.0.1_1/libexec /home/linuxbrew/.linuxbrew/Cellar/gaemr/1.0.1_1/bin/GAEMR.py --help
/usr/bin/env: ‘python’: No such file or directory
Error: brewsci/bio/gaemr: failed
<0> expected but was
<127>.
```

On MacOS: 
```
ImportError: No module named Bio
Error: brewsci/bio/gaemr: failed
<0> expected but was
<1>.
```

Ok, so we will ignore the error for Bio? I will try to fix python on linux then.

sjackman:
GAEMR hasn't been updated since 2012. I'm thinking that we should refocus our ORCA efforts on actively maintained packages. Let's remove GAEMR from ORCA and close this issue.

PR #263, opened 2018-05-02
sdsl-lite: import from brewsci/science

suujia:

-----

### DIFF
--- a/Formula/sdsl-lite.rb
+++ b/Formula/sdsl-lite.rb
@@ -20,7 +20,7 @@ class SdslLite < Formula
     system "./install.sh", prefix
     pkgshare.install "examples", "extras", "tutorial"
     
-     # Reduce memory usage for Circle CI.
+    # Reduce memory usage for Circle CI.
     ENV["MAKEFLAGS"] = "-j8" if ENV["CIRCLECI"]
   end
 
--- a/Formula/sdsl-lite.rb
+++ b/Formula/sdsl-lite.rb
@@ -16,11 +16,11 @@ class SdslLite < Formula
   needs :cxx11
 
   def install
+     # Reduce memory usage for Circle CI.
+    ENV["MAKEFLAGS"] = "-j8" if ENV["CIRCLECI"]
     ENV.cxx11
     system "./install.sh", prefix
     pkgshare.install "examples", "extras", "tutorial"
-    # Reduce memory usage for Circle CI.
-    ENV["MAKEFLAGS"] = "-j8" if ENV["CIRCLECI"]
   end
 
   test do
--- a/Formula/sdsl-lite.rb
+++ b/Formula/sdsl-lite.rb
@@ -10,6 +10,8 @@ class SdslLite < Formula
     :revision => "0546faf0552142f06ff4b201b671a5769dd007ad",
     :tag => "v2.1.1"
   revision 2
+  head "https://github.com/simongog/sdsl-lite.git"
+  
   depends_on "cmake" => :build
 
   needs :cxx11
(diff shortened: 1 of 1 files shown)


sjackman:
```
Error: No available formula with the name "libdivsufsort"
```
You'll need to first migrate `libdivsufsort`, which this formula depends on.

suujia:
Ok sounds good. This is taken from sdsl-lite.rb: 

```
 # this library is now part of SDSL - should remove the formula?
  conflicts_with "libdivsufsort"    
```

sjackman:
Oh, hah. In that case, do not migrate `libdivsufsort`, and instead remove
```ruby
  conflicts_with "libdivsufsort"    
```

suujia:
I have updated it. Can we run the tests again?

sjackman:
```
brewsci/bio/sdsl-lite:
  * New formulae should not have a HEAD spec
```
Although I like having the `head` URL, `brew audit` complains when it exists in a new formula, which prevents bottles being created, and we don't currently have a way to disable this test. So, for now, please remove the `head` line. We can add it back in after merging this PR.

sjackman:
```
brewsci/bio/sdsl-lite:
  * C: 13: col 1: Trailing whitespace detected.
```

sjackman:
CircleCI on Linux
```
[ 87%] Building CXX object lib/CMakeFiles/sdsl.dir/structure_tree.cpp.o
g++-5: internal compiler error: Killed (program cc1plus)
```
https://circleci.com/gh/brewsci/homebrew-bio/2501

This error happens when it uses too much memory on CircleCI on Linux, which has only 4 GB of memory. For the fix, see for example
https://github.com/brewsci/homebrew-bio/blob/c76935a3896354ab70354f15966d50175c84cc95/Formula/fwdpp.rb#L25-L26

Try first 16, if that fails, try 8, then 4, 2, 1, until you find one that works.

jonchang:
Looks like 8 is the magic number.

```
  * C: 22: col 1: Trailing whitespace detected.
  * C: 23: col 6: Incorrect indentation detected (column 5 instead of 4).
```

sjackman:
On both CircleCI for Linux and TravisCI.
```
==> brew test brewsci/bio/sdsl-lite --verbose
Testing brewsci/bio/sdsl-lite
Error: brewsci/bio/sdsl-lite: failed
The selected compiler doesn't support C++11: c_compiler
```
https://travis-ci.org/brewsci/homebrew-bio/builds/377023241#L2533

This error is odd. GCC 5 definitely supports C++11.

Formula/sdsl-lite.rb
```diff
@@ -0,0 +1,34 @@
...
+  depends_on "cmake" => :build
+
+  needs :cxx11
+
+  def install
+    ENV.cxx11
+    system "./install.sh", prefix
+    pkgshare.install "examples", "extras", "tutorial"
+    # Reduce memory usage for Circle CI.
+    ENV["MAKEFLAGS"] = "-j8" if ENV["CIRCLECI"]
```
sjackman: Setting `MAKEFLAGS` needs to happen before `install.sh` to have an effect.

sjackman:
```
The selected compiler doesn't support C++11: c_compiler
```

This error will be resolved by PR https://github.com/Linuxbrew/brew/pull/705 once it's merged.

sjackman:
Merged PR https://github.com/Linuxbrew/brew/pull/705

suujia:
Rebuild:  Circle CI passes, travis error: 
``` 
The job exceeded the maximum time limit for jobs, and has been terminated.
```

sjackman:
Merged. Thanks, Susie!

Issue #194, opened 2018-04-04
meraculous: Migrate


sjackman:
@aersoares81 wrote…

> Hi! I wanted to install Meraculous, but I thought it would be best to migrate the formula first.
First I edited the old formula because it points to an old version that's not available anymore, but now the formula breaks with the error "Error: No such file or directory @ rb_sysopen - src/c/CMakeLists.txt". Can anyone help a git newbie to fix that? I don't know why it's breaking since I just updated the URL/sha256.

https://github.com/brewsci/homebrew-bio/issues/15#issuecomment-378606794

PR #272, opened 2018-05-12
cytoscape: import from brewsci/science

suujia:

-----

### DIFF
--- a/Formula/cytoscape.rb
+++ b/Formula/cytoscape.rb
@@ -2,17 +2,16 @@ class Cytoscape < Formula
   # cite Saito_2012 "https://doi.org/10.1038/nmeth.2212"
   desc "Network Data Integration, Analysis, and Visualization in a Box"
   homepage "http://www.cytoscape.org/"
-
   url "https://github.com/cytoscape/cytoscape/archive/3.6.1.tar.gz"
-  sha256 "e20a04b031818005090bd65b78bb08813b7a8e018c73496d41f2f00014d6ae18"
+  sha256 "f172c9e459293661a83d2618da511901410336b76b0c1716697e60f821c2f150"
 
   depends_on :java => "1.8"
 
   def install
-    inreplace "cytoscape.sh", "$script_path", prefix
-    prefix.install %w[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
-    bin.install_symlink prefix/"cytoscape.sh" => "cytoscape"
-    (bin/"cytoscape").write_env_script libexec/"cytoscape", Language::Java.java_home_env("1.8")
+    ENV["JAVA_HOME"] = Formula["jdk"].prefix if ENV["CIRCLECI"]
+    inreplace "cy.sh", "$script_path", prefix
+    prefix.install %w[cy.sh apps gen_vmoptions.sh framework sampleData]
+    bin.install_symlink prefix/"cy.sh" => "cytoscape"
   end
 
   def caveats
--- a/Formula/cytoscape.rb
+++ b/Formula/cytoscape.rb
@@ -6,11 +6,11 @@ class Cytoscape < Formula
   url "http://chianti.ucsd.edu/cytoscape-3.4.0/cytoscape-3.4.0.tar.gz"
   sha256 "e20a04b031818005090bd65b78bb08813b7a8e018c73496d41f2f00014d6ae18"
 
-  depends_on :java => "1.8"
+  depends_on java => "1.8"
 
   def install
     inreplace "cytoscape.sh", "$script_path", prefix
-    prefix.install %W[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
+    prefix.install %w[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
     bin.install_symlink prefix/"cytoscape.sh" => "cytoscape"
   end
 
--- a/Formula/cytoscape.rb
+++ b/Formula/cytoscape.rb
@@ -10,7 +10,7 @@ class Cytoscape < Formula
 
   def install
     inreplace "cytoscape.sh", "$script_path", prefix
-    prefix.install %w[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
+    prefix.install %W[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
     bin.install_symlink prefix/"cytoscape.sh" => "cytoscape"
   end
 
--- a/Formula/cytoscape.rb
+++ b/Formula/cytoscape.rb
@@ -9,6 +9,8 @@ class Cytoscape < Formula
   depends_on :java => "1.8"
 
   def install
+    cmd = Language::Java.java_home_cmd("1.8")
+    ENV["JAVA_HOME"] = Utils.popen_read(cmd).chomp
     inreplace "cytoscape.sh", "$script_path", prefix
     prefix.install %w[cytoscape.sh apps gen_vmoptions.sh framework sampleData]
     bin.install_symlink prefix/"cytoscape.sh" => "cytoscape"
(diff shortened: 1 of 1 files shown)


sjackman:
@suujia Can you determine from their documentation or by experimentation whether Cytoscape works with Java 9 or Java 10, or only Java 8?

suujia:
``` 
Set the JAVA_HOME environment variable to the JDK 1.8 installation directory.
This is only necessary if you have multiple versions of Java installed - if JDK 1.8 
is the only one, Cytoscape will be able to automatically find it without the need 
for an environment variable. 
```

suujia:
So I've manually set the java_home environment variables. Then got this error message
```
==> Verifying cytoscape-3.4.0.tar.gz checksum
tar xzf /home/linuxbrew/.cache/Homebrew/cytoscape-3.4.0.tar.gz
Error: no implicit conversion of nil into String
```

sjackman:
My guess is that `Language::Java.java_home_cmd("1.8")` does not work on Linux.

```
$ brew irb <<<'Language::Java.java_home_cmd("1.8")'
==> Interactive Homebrew Shell
Example commands available with: brew irb --examples
Switch to inspect mode.
Language::Java.java_home_cmd("1.8")
nil
```
Yep.

Would you like to take a stab at fixing it?
See https://github.com/Linuxbrew/brew/blob/269be25411af00b5ecf9764a92c8061dc79ed8ca/Library/Homebrew/language/java.rb#L3

Alternatively, we may want to consider dropping from ORCA all tools that require specifically Java 8.
@suujia Could you please make a list of those tools?

suujia:
Hi, just wanted to get a clarification. ORCA doesn't specifically call brew install JDK@8 in its Dockerfiles. Where would tools that specifically require Java 8 appear? I thought brew install jdk installs only the latest version of JDK? Thanks :)

sjackman:
Correct `brew install jdk` install JDK 9, and JDK 8 is not currently installed. Any tools that require specifically JDK 8 are not currently installed in ORCA. One option is to continue to leave them out, if it's too troublesome to get JDK 8 working. I don't think many tools require specifically Java 8, so that may be a reasonable option.

```
cd $(brew --repo brewsci/bio)
grep -Fr '"1.8"' .
./Formula/nextflow.rb:  depends_on :java => "1.8"
./Formula/r8s.rb:  version "1.8"
```

suujia:
```
No such file or directory @ rb_sysopen - cytoscape.sh
```
related issue: #194 with rb_sysopen issue

sjackman:
```
Warning: Formula reports different SHA256: e20a04b031818005090bd65b78bb08813b7a8e018c73496d41f2f00014d6ae18
```
https://travis-ci.org/brewsci/homebrew-bio/builds/379974409#L1484

This error means that the upstream tarball has changed since this formula was last updated (which is not a nice thing to do). They've also renamed the shell script from `cytoscape.sh` to `cy.sh`.

suujia:
```
$ java -version
java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
```

```
$ which java
/home/linuxbrew/.linuxbrew/bin/java
```

```
$ namei /home/linuxbrew/.linuxbrew/bin/java

f: /home/linuxbrew/.linuxbrew/bin/java
 d /
 d home
 d linuxbrew
 d .linuxbrew
 d bin
 l java -> ../Cellar/jdk@8/1.8.0-151/bin/java
   d ..
   d Cellar
   d jdk@8
   d 1.8.0-151
   d bin
   - java
```

therefore (bottle): 
JAVA_HOME=/.linuxbrew/Cellar/jdk@8/
and corresponding jdk: 
JAVA_HOME=/.linuxbrew/Cellar/jdk/

whereas on macOS it's JAVA_HOME=$(/usr/libexec/java_home) 

*** change in java.rb to accept linux java_home settings and to look for jdk@8 in java_requirements

sjackman:
I didn't know about `namei`. That's super useful! Thanks, Susie!

sjackman:
I'm not sure where or if these functions in `language/java.rb` are actually used, oddly.
I'd suggest changing this bit of code to loop over `jdk` as well as `jdk@8`.
See https://github.com/Linuxbrew/brew/blob/master/Library/Homebrew/requirements/java_requirement.rb#L86-L91

suujia:
Java_home_env is used in a number of java 1.8 only formulas in homebrew/homebrew-core. An option is to allow linuxbrew formulas to call it with above changes.

sjackman:
> It looks like I have to do a link --force --overwrite separately even after setting the java_home and path?

Preferably without `brew link jdk@8` if possible. Though I'd be okay with shipping ORCA with `jdk@8` only (without `jdk` 9) and `brew link --force jdk@8` if necessary. I'm hoping that linking `jdk@8` is not necessary.

suujia:
After installing jdk on linuxbrew/linuxbrew docker image: 
```
$ brew irb
$ JavaRequirement.new(%w[1.7+]).send("possible_javas")
[#<Pathname:/home/linuxbrew/.linuxbrew/opt/jdk/bin/java>, #<Pathname:/home/linuxbrew/.linuxbrew/bin/java>]

$ JavaRequirement.new(%w[1.7+]).send("preferred_java")
=> #<Pathname:/home/linuxbrew/.linuxbrew/opt/jdk/bin/java>

$ JavaRequirement.new(%w[1.7+]).send("java_home_cmd")
=> undefined method `java_home_cmd' for #<JavaRequirement: "java" [] version="1.7+">

$ JavaRequirement.new(%w[1.8]).send("satisfied?")
=> false
$ JavaRequirement.new(%w[1.8+]).send("satisfied?")
=> true
```

After installing jdk@8  
```
JavaRequirement.new(%w[1.8]).send("satisfied?")
=> false
```

Making the previous comments changes to java_requirements.rb (after installing jdk@8): 
```
$ JavaRequirement.new(%w[1.7+]).send("possible_javas")
=> [#<Pathname:/home/linuxbrew/.linuxbrew/opt/jdk/bin/java>, #<Pathname:/home/linuxbrew/.linuxbrew/bin/java>, #<Pathname:/home/linuxbrew/.linuxbrew/opt/jdk@8/bin/java>, #<Pathname:/home/linuxbrew/.linuxbrew/bin/java>]

$ JavaRequirement.new(%w[1.8]).send("preferred_java")
=> #<Pathname:/home/linuxbrew/.linuxbrew/opt/jdk@8/bin/java>

$ JavaRequirement.new(%w[1.8]).send("satisfied?")
=> true
```

So possible_javas and preferred_java seems to be able to pick up jdk@8. So I then tested nextflow.rb which depends on `java 1.8` only 
```
$ brew test nextflow
Testing brewsci/bio/nextflow
WARN: Nextflow support for Java 9 is still in beta stage and some advanced features are not supported
WARN: It is suggested the use of Java version 8
Picked up _JAVA_OPTIONS:  -Duser.home=/home/linuxbrew/.cache/Homebrew/java_cache
```
Pretty odd, I thought it would be able to pick up jdk@8 after the change in preferred java hm. Even after setting the `PATH` and `JAVA_HOME` I am unable to get nextflow working without `brew link --force --overwrite`

sjackman:
Here's the logic that Nextflow uses to identify the version of Java:
https://github.com/nextflow-io/nextflow/blob/4b181df3976c404e3e68651f1b05488d1dd5954a/nextflow#L288-L307

`depends_on :java => "1.8"` only checks that the required version of Java exists. It doesn't unfortunately tell the installed software how to find that version of Java. To do that, you'd need to set either `JAVA_CMD` or `JAVA_HOME` or ensure that version occurs first in `PATH`.

sjackman:
Good progress though! Could you please open a pull request against Linuxbrew/brew with your changes to the Java search logic?

PR #288, opened 2018-05-28
bcalm: import from brewsci/science

sjackman:


sjackman:
CircleCI exceeded available memory.

| Jobs | Status |
|---|---|
| 32 | ❌ |
| 16 | |

PR #297, opened 2018-05-29
gatk 4.0.4.0

jonchang:


jonchang:
test failed on linux- probably java related

sjackman:
Yep. It seems that GATK requires Java 8 and does not work with Java 9. See also PR #158 gatk 40.2.1

PR #315, opened 2018-06-20
seq-gen: import from brewsci/science

jonchang:


jonchang:
Travis failed due to github api errors. Merging.

PR #316, opened 2018-06-20
beagle: migrate to homebrew/core

jonchang:
See https://github.com/Homebrew/homebrew-core/pull/29247

-----


jonchang:
Our first migration to homebrew/core! What an event.

PR #344, opened 2018-08-06
bali-phy 3.3

bredelings:
Also
 * add depends on cairo, so that drawtree is (automatically) built.
 * add build depends on pkg-config
 * add a real test.

Note that, as the author, I've been serving a Formula for bali-phy from https://github.com/bredelings/homebrew-bioinformatics

This formula causes a conflict in the command `brew install bali-phy` for users who have tapped both brewsci/bio and my repo.  This PR is an attempt to integrate my changes.  If this works, I can decommission my tap and recommend users to use this one instead, I suppose.

-----

### DIFF
--- a/Formula/bali-phy.rb
+++ b/Formula/bali-phy.rb
@@ -2,8 +2,8 @@ class BaliPhy < Formula
   # cite Suchard_2006: "http://dx.doi.org/10.1093/bioinformatics/btl175"
   desc "Bayesian co-estimation of phylogenies and multiple alignments"
   homepage "http://www.bali-phy.org/"
-  url "https://github.com/bredelings/BAli-Phy/archive/3.1.4.tar.gz"
-  sha256 "83124383533d12ea335e81e63a202ca97d850e8962b1a77e6166ae0fd0ebffa0"
+  url "https://github.com/bredelings/BAli-Phy/archive/3.3.tar.gz"
+  sha256 "84132dd0ec8a62eb6697374c62847e4b84082cbf09977f0e157e8407db010ccc"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-bio"
@@ -14,6 +14,8 @@ class BaliPhy < Formula
   depends_on "meson" => :build
   depends_on "ninja" => :build
   depends_on "pandoc" => :build
+  depends_on "pkg-config" => :build
+  depends_on "cairo"
 
   def install
     flags = %w[-C build install]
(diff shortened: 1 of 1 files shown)


bredelings:
The travis-ci failure seems to be because pandoc doesn't build.

jonchang:
Hi @bredelings, thanks for contributing to Brewsci! It was good to meet you in Columbus.

```
* C: 17: col 3: dependency "pandoc" (line 17) should be put before dependency "pkg-config" (line 16)
```

We needed to rebuild bottles due to https://github.com/Linuxbrew/brew/pull/773, so I made this change to trigger that build.

> add depends on cairo, so that drawtree is (automatically) built.

I didn't originally include this since on Linux this brings in all of x11. But it's not a big deal. We can change this to a `:recommended` dependency if there's a need for it.

> This formula causes a conflict in the command brew install bali-phy for users who have tapped both brewsci/bio and my repo. This PR is an attempt to integrate my changes. If this works, I can decommission my tap and recommend users to use this one instead, I suppose.

Yes, that's fine. If you want to continue to maintain your own tap the `brew tap-pin` command can be useful. Otherwise, you can specify a [tap_migrations.json](https://github.com/brewsci/homebrew-bio/blob/master/tap_migrations.json) file in your repository so that your current users can migrate to this one.

bredelings:
Cool!  Yes, good to meet you in Columbus!

Yeah, it is kind of unfortunate that cairo pulls in all of X, when I just need the pdf-generation facilities.  I don't want to deliver partially-working software on mac though, and hopefully bottles will speed things up.  If this causes problems then a :recommends dependency is fine.

Thanks for mentioning `brew tap-pin` and `tap_migrations.json`.  I didn't know about these.  I think I'll switch the installation instructions to mention brewsci and if everything works well for a few releases I'll also add a tap_migrations.json.  Let's see how it goes.

jonchang:
Testing to see if https://github.com/Linuxbrew/homebrew-test-bot/pull/100 fixes bottling...

sjackman:
For future PRs, the commit log message ought to be `bali-phy 3.3 (new formula)`

sjackman:
I've rebased this PR onto origin/master to trigger a rebuild. I also squashed the two commits into one commit and changed the commit log message as above.

sjackman:
Merged. Thanks for your first contribution to Brewsci/bio, Benjamin! It's always a fine thing to have the upstream author on board with packaging. 🏅

PR #354, opened 2018-08-09
quast 5.0.0

alexeigurevich:

-----

### DIFF
--- a/Formula/quast.rb
+++ b/Formula/quast.rb
@@ -4,7 +4,7 @@ class Quast < Formula
   # cite Mikheenko_2016: "https://doi.org/10.1093/bioinformatics/btw379"
   # cite Mikheenko_2018: "https://doi.org/10.1093/bioinformatics/bty266"
   desc "QUAST: Quality Assessment Tool for Genome Assemblies"
-  homepage "http://quast.sourceforge.net/"
+  homepage "https://quast.sourceforge.io/"
   url "https://downloads.sourceforge.net/project/quast/quast-5.0.0.tar.gz"
   sha256 "129013967279e9bf501be67e2e59c3b8eed1de82b19163cb528aa618a21b8d73"
   head "https://github.com/ablab/quast.git"
--- a/Formula/quast.rb
+++ b/Formula/quast.rb
@@ -16,6 +16,7 @@ class Quast < Formula
   end
 
   depends_on "zlib" unless OS.mac?
+  depends_on "python"
 
   def install
     prefix.install Dir["*"]
@@ -23,6 +24,11 @@ class Quast < Formula
       "quast.py" => "quast", "metaquast.py" => "metaquast", "quast-lg.py" => "quast-lg", "icarus.py" => "icarus"
     # Compile the bundled aligner so that `brew test quast` does not fail.
     system "#{bin}/quast", "--test"
+    if OS.linux?
+      inreplace "#{bin}/../quast.py", "#!/usr/bin/env python", "#!/usr/bin/env python3"
+      inreplace "#{bin}/../metaquast.py", "#!/usr/bin/env python", "#!/usr/bin/env python3"
+      inreplace "#{bin}/../icarus.py", "#!/usr/bin/env python", "#!/usr/bin/env python3"
+    end
   end
 
   test do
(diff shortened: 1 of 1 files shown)


alexeigurevich:
Hi, Shaun @sjackman !

Could you assist with updating Quast to the new major release?
I see that the problem in the failed Linux builds is
`/usr/bin/env: ‘python’: No such file or directory`
when the system tries to execute `./quast.py --test` (and quast.py indeed starts with `#!/usr/bin/env python`).
However, I do not understand how this should be properly fixed. Note that my formula is based on existing (and working) quast-4.6.3 formula and `./quast.py` file always had `#!/usr/bin/env python` in the beginning.

Thanks,
Alexey

sjackman:
Hi, Alexei! Yeah, I've been waiting the release of QUAST 5 with anticipation!

Let's use Python 3 rather than Python 2. Please add `depends_on "python"` to depend on Python 3 (currently Python 3.7), and change the shebang line to `#!/usr/bin/env python3`.

macOS is using the host's `/usr/bin/python` (2.7.10), which is why it succeeds. On Linux, we don't assume the host has any Python interpreter installed.

Thanks for this PR!

alexeigurevich:
Ok, got it. Is it possible to specify some "patch" to be applied after unpacking Quast `.tar.gz` archive? Or should we release something like 5.0.1 with the modified shebang? We can't change anything in already uploaded 5.0.0.tar.gz. 

Moreover, I am not sure that releasing a general `.tar.gz` with explicitly specified `python3` is a perfect solution. Since a user may have python2 only, too. Do you know whether there is a universal solution?

sjackman:
You can patch the `quast` executable after installation. Add…
```ruby
inreplace bin/"quast", "#!/usr/bin/env python", "#!/usr/bin/env python3
```

alexeigurevich:
I slightly modified your suggestion (otherwise it was not working due to symlinking issue). And also specified this `inreplace` for Linux only. Please take a look whether you agree with my version.

sjackman:
I'd prefer to use Python 3 on both macOS and Linux, for consistency on both platforms. Sound good?

Formula/quast.rb
```diff
@@ -14,20 +15,20 @@ class Quast < Formula
...
-    # Remove the precompiled E-MEM binary.
-    rm_f "quast_libs/MUMmer/e-mem-osx"
     prefix.install Dir["*"]
-    bin.install_symlink "../quast.py", "../metaquast.py", "../icarus.py",
-      "quast.py" => "quast", "metaquast.py" => "metaquast", "icarus.py" => "icarus"
+    bin.install_symlink "../quast.py", "../metaquast.py", "../quast-lg.py", "../icarus.py",
+      "quast.py" => "quast", "metaquast.py" => "metaquast", "quast-lg.py" => "quast-lg", "icarus.py" => "icarus"
     # Compile the bundled aligner so that `brew test quast` does not fail.
     system "#{bin}/quast", "--test"
+    if OS.linux?
```
sjackman: Please move the `inreplace` above `system "#{bin}/quast", "--test"` and add a blank line before and after.

alexeigurevich: Yes, sure. Sorry, I am testing `brew install --build-from-source` on Mac, so it is difficult to catch everything.

sjackman: No worries at all, Alexei.

alexeigurevich:
>I'd prefer to use Python 3 on both macOS and Linux, for consistency on both platforms. Sound good?

Yes, that makes sense. However, I tried to specify 
`  depends_on "python3"` in the formula for consistency but `brew audit` told me that this is not good and I should use simply `"python"` instead (which probably means python2 on Mac). Should I ignore `brew audit`?

sjackman:
`depends_on "python"` is Python 3 on both Linux and macOS. On macOS if you want to use the host's Python interpreter, you don't add any dependency at all. The dependency on the host OS is implied.

sjackman:
Is `quast-lg` a synonym for `quast --large`?

sjackman:
`brew postinstall python` failed on TravisCI, but that's unrelated to this PR, and not critical.
https://travis-ci.org/brewsci/homebrew-bio/builds/414264629#L3210

sjackman:
Merged! Thanks, Alexei!

alexeigurevich:
> Is quast-lg a synonym for quast --large?
Yes, exactly. We added it just in case and mostly for new users who will start familiarize themselves with Quast from Quast-LG.

sjackman:
No worries at all! My pleasure.

PR #378, opened 2018-08-20
busco: compatibility with python 3.7

suujia:

-----

### DIFF
--- a/Formula/busco.rb
+++ b/Formula/busco.rb
@@ -2,9 +2,10 @@ class Busco < Formula
   # cite Waterhouse_2017: "https://doi.org/10.1093/molbev/msx319"
   # cite Sim_o_2015: "https://doi.org/10.1093/bioinformatics/btv351"
   desc "Assess genome assembly completeness with single-copy orthologs"
-  homepage "http://busco.ezlab.org"
+  homepage "https://busco.ezlab.org"
   url "https://gitlab.com/ezlab/busco/repository/3.0.2/archive.tar.bz2"
   sha256 "cd0699545a126c7cc94604eef7c8dc50379b5d11becbad3a0f55d995a4c5e1c0"
+  revision 1
   head "https://gitlab.com/ezlab/busco.git"
 
   bottle do
@@ -14,7 +15,7 @@ class Busco < Formula
     sha256 "d5fc62e2ab2d5cf248d6cac19de3be8c69b6ec4ebbcd8d132974dbf0add3d6ca" => :x86_64_linux
   end
 
-  depends_on "python3"
+  depends_on "python"
   depends_on "blast" => :recommended
   # Also depends on augustus and hmmer
 


Formula/busco.rb
```diff
@@ -19,8 +19,11 @@ class Busco < Formula
   # Also depends on augustus and hmmer
 
   def install
+    xy = Language::Python.major_minor_version "python3"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python${xy}/site-packages"
+
     inreplace Dir["scripts/*.py"], "#!/usr/bin/env python", "#!#{HOMEBREW_PREFIX}/bin/python3"
-    system "python3", "setup.py", "install", "--prefix=#{prefix}"
+    system "python", "setup.py", "install", "--user", "--prefix="
```
sjackman: Leave this line as `python3` and change…
```diff
-depends_on "python3"
+depends_on "python"
```

sjackman: I believe `--prefix=…` and `--user` are mutually exclusive.

suujia:
https://gitlab.com/ezlab/busco/issues/27
I made the following changes: 

1) 
```
I have never faced the problem myself, but what about 
python setup.py install --user --prefix= as I suggested. (yes empty prefix)
```
2) 
```
Another possible solution is to append the path to your globally installed BUSCO modules to PYTHONPATH.
```
When I looked in orca image location for site-packages, I found it here and added it to path: 
`export PYTHONPATH="${PYTHONPATH}:/.linuxbrew/Cellar/busco/3.0.2/lib/python3.6/site-packages"`
Even though the python that was downloaded in orca is python 3.7, so I think BUSCO couldn't find pipebricks.

Formula/busco.rb
```diff
@@ -14,13 +14,17 @@ class Busco < Formula
...
   # Also depends on augustus and hmmer
 
   def install
+    xy = Language::Python.major_minor_version "python3"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python${xy}/site-packages"
+    system python, "import pipebricks"
+
     inreplace Dir["scripts/*.py"], "#!/usr/bin/env python", "#!#{HOMEBREW_PREFIX}/bin/python3"
-    system "python3", "setup.py", "install", "--prefix=#{prefix}"
+    system "python3", "setup.py", "install", "--user", "--prefix="
```
sjackman: I believe `--user` will install somewhere in the user's home directory. We want to use `--prefix=#{prefix}` to install in a directory that we specify.

suujia:
The changes worked! 😺

Formula/busco.rb
```diff
@@ -14,11 +14,14 @@ class Busco < Formula
...
-  depends_on "python3"
+  depends_on "python"
   depends_on "blast" => :recommended
   # Also depends on augustus and hmmer
 
   def install
+    xy = Language::Python.major_minor_version "python3"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python${xy}/site-packages"
+
     inreplace Dir["scripts/*.py"], "#!/usr/bin/env python", "#!#{HOMEBREW_PREFIX}/bin/python3"
```
sjackman: Is this changed needed, and does it work without it?

sjackman:
Woohoo!

suujia:
Yup, removing the PYTHONPATH command words as well! Thought that was what's breaking it but ok! :)

sjackman:
Please add `revision 1` after `sha256`.

suujia:
this should be good to go :)

sjackman:
Merged. Thanks again, Susie! Have a good weekend! 🌞

PR #380, opened 2018-08-20
smalt: import from brewsci/science

suujia:


suujia:
Thanks @tseemann ! :)

PR #388, opened 2018-08-23
phyx 0.999

jonchang:


sjackman:
```
failed: brew test brewsci/bio/phyx --verbose
Testing brewsci/bio/phyx
/home/linuxbrew/.linuxbrew/Cellar/phyx/0.999/bin/pxseqgen -t /home/linuxbrew/.linuxbrew/Cellar/phyx/0.999/share/phyx/pxseqgen_example/seqgen_test.tre -o output.fa
Error: brewsci/bio/phyx: failed
execution expired
```
This means that the test took longer than 5 minutes. Is there a simpler test that could be run instead?

jonchang:
@sjackman that test should be more-or-less instant, so no clue what's going on. Couldn't repro on my local machine.

jonchang:
And travis didn't have an issue with it (1 second). Restarting circle to see if it's transient.

sjackman:
If it happens again on CircleCI, you can ssh into the CircleCI machine to troubleshoot interactively.

jonchang:
Not sure why `version_scheme` was necessary to get test-bot to build bottles rather than merely test them. `echo 'Version.new("0.99") < Version.new("0.999")' | brew irb` returns true.

sjackman:
Yeah, `version_scheme` shouldn't be necessary.

sjackman:
Superseded by https://github.com/brewsci/homebrew-bio/pull/436

PR #412, opened 2018-08-31
tbl2asn: revision for sha256 update and Linux fix

jonchang:

-----

* the sha256 hash changed for both Mac and Linux due to a filename change of the compressed binary
* Linux version links against `libidn.so.11` but we provide `libidn.so.12`. I tested it and it doesn't seem to break anything so `patchelf` is used to fix up the linkage.

### DIFF
--- a/Formula/tbl2asn.rb
+++ b/Formula/tbl2asn.rb
@@ -37,8 +37,10 @@ class Tbl2asn < Formula
       system "patchelf",
         "--set-interpreter", HOMEBREW_PREFIX/"lib/ld.so",
         "--set-rpath", HOMEBREW_PREFIX/"lib",
-        "--replace-needed", "libidn.so.11", "libidn.so.12",
         bin/"tbl2asn"
+        # Normally we would use patchelf to make this change but it seems
+        # broken for this use-case. Use the Stick of Correction instead.
+        inreplace bin/"tbl2asn", "libidn.so.11", "libidn.so.12"
     end
     doc.install resource("doc")
   end
--- a/Formula/tbl2asn.rb
+++ b/Formula/tbl2asn.rb
@@ -6,11 +6,12 @@ class Tbl2asn < Formula
   version "25.6"
   if OS.mac?
     url "https://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/tbl2asn/mac.tbl2asn.gz"
-    sha256 "c79416ff5fea23baf4ac10ff1a67f7f6e099980a45ac878f649821ba7b68788b"
+    sha256 "f0ef9038bb021fa0c8b546a78539fe2f34c3b22c28986c3f99d216073366ff08"
   elsif OS.linux?
     url "https://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/tbl2asn/linux64.tbl2asn.gz"
-    sha256 "5306321c1e9cd709c41a47a01c8193cff20bc2c71141037e739dd8b59cb30dc2"
+    sha256 "f0e7d2fc6c68b39ff7b6abb1475e0a9d63c60887cfbd2857ccf0a0923eb4b96e"
   end
+  revision 1
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-bio"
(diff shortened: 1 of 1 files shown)


jonchang:
macOS+Linux:

```
brewsci/bio/tbl2asn:
  * stable: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed.
```

Linux:

```
Inconsistency detected by ld.so: dl-version.c: 224: _dl_check_map_versions: Assertion `needed != NULL' failed!
Error: brewsci/bio/tbl2asn: failed
<0> expected but was
<127>.
```

Not sure what's going on. I think `patchelf` is modifying the binary incorrectly or something. If I do `--replace-needed libidn.so.11 libidn.so.11 tbl2asn` I get the same error locally.

sjackman:
Would you like to look into building `tbl2asn` from source?

jonchang:
I did find it on the ftp site (it's at ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/CURRENT/ncbi.tar.gz) but the source for `tbl2asn.c` lists its version as 25.3 but the binaries provided by NCBI are 25.6. I'm going to hold off on building from source because I think NCBI gets grouchy when you submit to Genbank with old tools. I'll probably write them at some point asking about the version difference but I think for now using the binaries is our best bet.

Formula/tbl2asn.rb
```diff
@@ -31,14 +32,15 @@ class Tbl2asn < Formula
...
-      bin.install "linux64.tbl2asn" => "tbl2asn"
+    bin.install Dir["tbl2asn*"].first => "tbl2asn"
+    if OS.linux?
       system "patchelf",
         "--set-interpreter", HOMEBREW_PREFIX/"lib/ld.so",
         "--set-rpath", HOMEBREW_PREFIX/"lib",
         bin/"tbl2asn"
+        # Normally we would use patchelf to make this change but it seems
+        # broken for this use-case. Use the Stick of Correction instead.
+        inreplace bin/"tbl2asn", "libidn.so.11", "libidn.so.12"
```
sjackman: 😮 Wow. Amazed this works. I'll tuck that away in my back pocket.
Running `patchelf` a second time with `--replace-needed` may also work.

sjackman:
```
  * stable: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed.
```
We don't currently have the ability to override a failed audit, and the bottles don't transfer with a failed audit.

Two possible hack workarounds.
1. Commit the `url` change in one PR and delete the bottle, then in a second PR build the bottles.
2. Modify the `.circleci/config.yml` to add `brew test-bot … || true` to force it to be successful in a separate commit, merge the PR, and remove that commit before pushing.

I have a mild preference for option 2.

Formula/tbl2asn.rb
jonchang: Nope, tried that. I think there's a bug in patchelf with respect to how `tbl2asn` is linked.

jonchang:
Looks like NCBI updated its tbl2asn binaries again. Looking at the binary diff there appear to be substantial changes but I can't see anything different in the source code itself. Possibly in one of the header libraries. I'm going to hold off on updating this until I can find out a contact email at NCBI about WTF is going on.

sjackman:
@jonchang Any update on this PR?

Formula/tbl2asn.rb
tseemann: The old binary patch! Nice one. 

The last time I did that was to patch the Roche 454 Newbler assembler to use our Ion Torrent mate pair linker sequence instead.

jonchang:
Yeah I haven't talked to NCBI yet, the move has really taken a lot out of me. Will follow up on this next week.

Formula/tbl2asn.rb
sjackman: Amazing! That's a good story!

sjackman:
I wish that NCBI would bump the version number printed by `--help` when they update the `tbl2asn` executable. My copy of `tbl2asn` is now failing with this message…
```
[NULL_Caption] This copy of tbl2asn is more than a year old.  Please download the current version.
```
When I download the new version from the web site, it still reports version `25.6`, but this warning goes away.

sjackman:
I'm going to see if I can get this PR through `brew audit` in PR https://github.com/brewsci/homebrew-bio/pull/541

sjackman:
Superseded by PR https://github.com/brewsci/homebrew-bio/pull/541.

PR #421, opened 2018-09-05
idba: Bump revision for gcc on macOS

sjackman:
Fix the error:
```
dyld: Library not loaded: /usr/local/opt/gcc/lib/gcc/7/libgomp.1.dylib
  Referenced from: /usr/local/bin/idba_ud
  Reason: image not found
```


sjackman:
```
  * Non-executables were installed to "/usr/local/opt/idba/bin"
    The offending files are:
      /usr/local/opt/idba/bin/idba.dSYM
…
```
https://circleci.com/gh/brewsci/homebrew-bio/4002

Removing these `.dSYM` files, though I'm not sure why these files are being created.

PR #424, opened 2018-09-06
CircleCI: Set HOMEBREW_MAKE_JOBS=4

sjackman:
Reduce the number of default number of make jobs to 8.

The default number of HOMEBREW_MAKE_JOBS on CircleCI for Linux is 32.
The resources available reported by CircleCI are 2 CPU and 4 GB RAM.
32 processes of GCC often exceeds the 4 GB of available memory.
Since our instance has only 2 CPU, it's likely not any faster either.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ jobs:
       CIRCLE_REPOSITORY_URL: https://github.com/brewsci/homebrew-bio
       HOMEBREW_DEVELOPER: 1
       HOMEBREW_NO_AUTO_UPDATE: 1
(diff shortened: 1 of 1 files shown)


sjackman:
Counts of formulae that require fewer than 32 jobs.
```
   1     ENV["MAKEFLAGS"] = "-j16" if ENV["CIRCLECI"]
   8     ENV["MAKEFLAGS"] = "-j8" if ENV["CIRCLECI"]
   6     ENV["MAKEFLAGS"] = "-j4" if ENV["CIRCLECI"]
```

PR #455, opened 2018-10-03
flowcraft: Fix sha256

sjackman:
See https://github.com/assemblerflow/flowcraft/issues/134


sjackman:
```
  * stable: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed.
```
Grumble.

PR #457, opened 2018-10-03
nxrepair: Fix for Linux

sjackman:
Build matplotlib from source.


sjackman:
@tseemann Woohoo! I fixed the `Missing libraries: libz-a147dcb0.so.1.2.3` issue by building `matplotlib` from source rather than using the precompiled binary wheel package. It was as easy as adding `pip install matplotlib --no-binary=matplotlib`.

PR #502, opened 2018-10-25
canu 1.8

sjackman:
See https://github.com/marbl/canu/releases/tag/v1.8


sjackman:
```
==> brew install gcc
Error: SIGTERM
…
Too long with no output (exceeded 30m0s)
```
Well that's weird. Trying again.

PR #516, opened 2018-11-06
minimap2 2.14

sjackman:
See https://github.com/lh3/minimap2/releases/tag/v2.14


jonchang:
Looks like the Homebrew/brew remote works (I'm assuming that's the docker image change)!

sjackman:
Yep! And it nearly worked. `brew test-bot` oddly did not attempt to build bottles for Linux. I'll look into it.

PR #517, opened 2018-11-14
adam 0.25.0

heuermh:
Created with `brew bump-formula-pr`.

PR #518, opened 2018-11-14
adam 0.25.0

sjackman:
Fix a bug in the CircleCI configuration.
See PR #517.


sjackman:
Merged. Thank you for your continued contribution to Brewsci/bio, Michael! 🏅

PR #522, opened 2018-11-19
cannoli 0.3.0

heuermh:


sjackman:
CircleCi on Linux
```
==> /home/linuxbrew/.linuxbrew/Cellar/cannoli/0.3.0/bin/cannoli-submit --help
/home/linuxbrew/.linuxbrew/bin/spark-submit: line 2: /usr/libexec/java_home: No such file or directory
Picked up _JAVA_OPTIONS:  -Duser.home=/home/linuxbrew/.cache/Homebrew/java_cache
/home/linuxbrew/.linuxbrew/Cellar/apache-spark/2.3.2/libexec/bin/spark-class: line 99: /tmp/cannoli-test-20181119-10419-6ukrdt/bin/java: No such file or directory
Error: brewsci/bio/cannoli: failed
An exception occurred within a child process:
  Test::Unit::AssertionFailedError: <0> expected but was
<127>.
```
https://circleci.com/gh/brewsci/homebrew-bio/4892

`/usr/libexec/java_home` is a macOS utility, which doesn't exist and shouldn't be used on Linux. I'll look into why `spark-submit` is attempting to run `/usr/libexec/java_home`.

heuermh:
> Odd that the adam build in #518 didn't run into this, since the submit scripts should be similar if not identical s/adam/cannoli/;

I'm sorry, that is not true, they have diverged quite a bit, and what are the chances I can remember why. . .

sjackman:
Brewsci/bio has recently migrated its CircleCI configuration to use Homebrew/brew rather than Linuxbrew/brew to build bottles on Linux. Homebrew/brew now runs on both macOS and Linux! It's possible that it's related to this recent change.

sjackman:
The bug appears to be in `apache-spark`.
This line in the `apache-spark` formula
```ruby
    bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env("1.8"))
```
creates this line in the shell script `spark-submit`
```sh
JAVA_HOME="$(/usr/libexec/java_home --version 1.8)" exec "/home/linuxbrew/.linuxbrew/Cellar/apache-spark/2.3.2/libexec/bin/spark-submit" "$@"
```
Since `/usr/libexec/java_home` doesn't exist on Linux, it has the effect of setting `JAVA_HOME` to the empty string `JAVA_HOME=""`, which breaks the script `spark-class`, causing it to look for a Java interpreter at `$HOME/bin/java`, which doesn't exist. If `JAVA_HOME` were not set, it would look for the Java interpreter in the `PATH`.

This function `Language::Java::java_home_cmd` in Homebrew/brew needs to be fixed for Linux.
https://github.com/Homebrew/brew/blob/72411560803d854256a2f75d09409bc5235b3755/Library/Homebrew/language/java.rb#L3

sjackman:
I've opened PR https://github.com/Homebrew/brew/pull/5333 to resolve this issue.

heuermh:
Thank you, @sjackman!  I admire the patience you displayed on the Homebrew PR.  :)

PR #526, opened 2018-11-29
biobloomtools 2.2.0

JustinChu:

-----

### DIFF
--- a/Formula/biobloomtools.rb
+++ b/Formula/biobloomtools.rb
@@ -2,9 +2,8 @@ class Biobloomtools < Formula
   # cite Chu_2014: "https://doi.org/10.1093/bioinformatics/btu558"
   desc "BioBloom Tools (BBT): Bloom filter for bioinformatics"
   homepage "http://www.bcgsc.ca/platform/bioinfo/software/biobloomtools/"
-  url "https://github.com/bcgsc/biobloom/releases/download/2.0.13/biobloomtools-2.0.13.tar.gz"
-  sha256 "7b4aeef70feb3fc31db2f4695159523272eadd8787b33c2902e2970a7d569ba3"
-  revision 1
+  url "https://github.com/bcgsc/biobloom/releases/download/2.2.0/biobloomtools-2.2.0.tar.gz"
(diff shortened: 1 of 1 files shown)


JustinChu:
@sjackman I've fixed some issues but the logs look like an issue with sdsl-lite.

sjackman:
CircleCI on macOS
```
g++-8 -Wall -Wextra -Werror  -fopenmp -g -O2 -fopenmp  -o biobloommimaker biobloommimaker-BioBloomMIMaker.o biobloommimaker-Options.o ../Common/libcommon.a -lz -lsdsl -ldl 
Undefined symbols for architecture x86_64:
  "sdsl::util::demangle2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)", referenced from:
      std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > sdsl::util::class_name<unsigned long long>(unsigned long long const&) in biobloommimaker-BioBloomMIMaker.o
```
https://circleci.com/gh/brewsci/homebrew-bio/4918

One possibility is that it's an error caused by `sdsl-lite` being compiled with clang on macOS, whereas you're compiling Biobloomtools using GCC. You can likely reproduce this error locally on a macOS machine for troubleshooting.

sjackman:
> I've tested installing the code on mac os before. So this is a bit strange. How do you prevent it from installing sdsl with clang?

It's not easy but possible. Let's troubleshoot first. Try installing SDSL using Homebrew on macOS, and then `brew install -dv biobloomtools`.

> Also aren't I just using a header anyway?

We tested on Linux, and yes. It's possible (though strange) that the situation is different on macOS.

sjackman:
`sdsl-lite` on macOS is compiled with `clang`. You'll need to compile it with GCC.
See https://github.com/brewsci/homebrew-bio/blob/master/Formula/lofreq.rb for an example of a resource block. You'll need to remove the `sdsl-lite` dependency and build `sdsl-lite` in your formula using `gcc`. This change only applies to macOS. On Linux you could keep the `sdsl-lite` dependency.

PR #541, opened 2018-12-20
tbl2asn: Update sha256

sjackman:


sjackman:
Merged.
@jonchang Thanks for the `inreplace bin/"tbl2asn", "libidn.so.11", "libidn.so.12"` hack!

PR #569, opened 2019-03-11
bracken 2.2 (new formula)

gaberoo:
- [ x ] Does your build pass `brew audit --strict FORMULA` (after doing `brew install FORMULA`)?

-----

The test is a bit inelegant because there is no official `-h` flag. Rather, usage is displayed when arguments aren't recognised. So I redirect stderr to /dev/null to suppress the warning.

### DIFF
--- a/Formula/bracken.rb
+++ b/Formula/bracken.rb
@@ -1,0 +1,25 @@
+class Bracken < Formula
+  # cite Lu_2017: "https://doi.org/10.7717/peerj-cs.104"
+  desc "Bayesian estimation species abundances from Kraken output"
+  homepage "https://ccb.jhu.edu/software/bracken/index.shtml"
+  url "https://github.com/jenniferlu717/Bracken/archive/v2.2.tar.gz"
+  sha256 "307495060d3575c3a90c87cf82074d4a63d20cf4062658f161951c60a66fd319"
+
+  depends_on "gcc" if OS.mac?
+
+  fails_with :clang # uses OpenMP
+
+  def install
+    system "make", "-C", "src"
+
+    inreplace "bracken", "$DIR/src", libexec
+    inreplace "bracken-build", "$DIR/src", libexec
+
+    bin.install "bracken", "bracken-build"
+    libexec.install "src/est_abundance.py", "src/generate_kmer_distribution.py", "src/kmer2read_distr", "src/kreport2mpa.py"
+  end
+
+  test do
+    assert_match "Usage", shell_output("#{bin}/bracken -h 2> /dev/null")
+  end
+end


Formula/bracken.rb
```diff
@@ -0,0 +1,28 @@
...
+  url "https://github.com/jenniferlu717/Bracken/archive/v2.2.tar.gz"
+  sha256 "307495060d3575c3a90c87cf82074d4a63d20cf4062658f161951c60a66fd319"
+
+  depends_on "gcc" if OS.mac?
+
+  fails_with :clang # uses OpenMP
+
+  def install
+    cd "src" do
+      inreplace "Makefile", "g++", "$(CXX)"
```
sjackman:
```suggestion
      inreplace "Makefile", "g++", ENV.cxx
```
Or you can delete this line altogether. It's handled by a feature of Homebrew called Superenv.

Formula/bracken.rb
```diff
@@ -0,0 +1,28 @@
...
+      system "make"
```
sjackman: If you delete the above line, you could instead use
```suggestion
      system "make", "-C", "src"
```

Formula/bracken.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    inreplace "bracken", "$DIR/src", libexec
+    inreplace "bracken-build", "$DIR/src", libexec
+
+    bin.install "bracken", "bracken-build"
+    libexec.install "src/est_abundance.py", "src/generate_kmer_distribution.py", "src/kmer2read_distr", "src/kreport2mpa.py"
+  end
+
+  test do
+    assert_match "Usage", shell_output("#{bin}/bracken -h 2> /dev/null")
```
sjackman:
```suggestion
    assert_match "Usage", shell_output("#{bin}/bracken 2>&1")
```
Does removing the `-h` option work?

Formula/bracken.rb
gaberoo: Ah, interesting. For some reason I thought that my `brew install --interactive` wasn't working right.

sjackman: Use `which -a g++` and you should see that the first hit is a Homebrew shim script.

Formula/bracken.rb
gaberoo: No, unfortunately that's "worse":

```
 ❯ bracken
 >> Checking for Valid Options...
 ERROR: /database100mers.kmer_distrib does not exist
        Run bracken-build to generate the kmer distribution file.
```

versus

```
 ❯ bracken -h
/usr/local/bin/bracken: illegal option -- h
Usage: bracken -d MY_DB -i INPUT -o OUTPUT -r READ_LEN -l LEVEL -t THRESHOLD
  MY_DB          location of Kraken database
  INPUT          Kraken REPORT file to use for abundance estimation
  OUTPUT         file name for Bracken default output
  READ_LEN       read length to get all classifications for (default: 100)
  LEVEL          level to estimate abundance at [options: D,P,C,O,F,G,S] (default: S)
  THRESHOLD      number of reads required PRIOR to abundance estimation to perform reestimation (default: 0)
```

Altogether this doesn't make that much of a difference, but just looks like somewhat of a "hack".

Formula/bracken.rb
gaberoo: Done.

Actually, I only just realised now that I get a install warning on macOS. I was primarily focusing on Linux.

```
 ❯ brew install bracken
==> Installing bracken from brewsci/bio
==> Downloading https://github.com/jenniferlu717/Bracken/archive/v2.2.tar.gz
Already downloaded: /Users/gabriel/Library/Caches/Homebrew/downloads/8582634cc4e28766810162dd38e7eb8a96469eb06b9e92332c5a7cddfd043f4f--Bracken-2.2.tar.gz
==> make
Warning: brewsci/bio/bracken dependency gmp was built with a different C++ standard
library (libc++ from clang). This may cause problems at runtime.
🍺  /usr/local/Cellar/bracken/2.2: 10 files, 155.5KB, built in 7 seconds
>>> elapsed time 14s
```

Any suggestion?

sjackman: You can safely ignore this warning.

sjackman:
Merged. Thanks again, Gabriel! 🏆
https://twitter.com/brew_sci/status/1105141654707556352

PR #572, opened 2019-03-15
DAS_Tool 1.1.1 (new formula)

gaberoo:

-----

I've been trying out a bunch of new software, and have been bummed that they aren't in Homebrew, so there are quite a few formula in the pipeline.

For this one I'm not sure what the policy is. It uses R for some plotting. I've inspired myself from some of the packages that have Python dependencies and installed all R dependencies in a library that is local to this package only. Is this the way it should be done?

I also assume there will be quite some style edits. I'm happy to follow your lead.

### DIFF
--- a/Formula/das_tool.rb
+++ b/Formula/das_tool.rb
@@ -1,0 +1,40 @@
+class DasTool < Formula
+  # cite Sieber_2018: "https://doi.org/10.1038/s41564-018-0171-1"
+  desc "Genomic binning refiner"
+  homepage "https://github.com/cmks/DAS_Tool"
+  url "https://github.com/cmks/DAS_Tool/archive/1.1.1.tar.gz"
+  sha256 "2a55f67b5331251d8fd5adea867cc341363fbf7fa7ed5c3ce9c7679d8039f03a"
+  head "https://github.com/cmks/DAS_Tool.git"
+
+  if OS.mac?
+    depends_on "gettext"
+  else
+    depends_on "icu4c"
+    depends_on "unzip" => :build
+  end
+
+  depends_on "diamond"
+  depends_on "prodigal"
+  depends_on "pullseq"
+  depends_on "r"
+
+  def install
+    mkdir_p buildpath/"lib/R"
+    ENV["R_LIBS_SITE"] = "#{buildpath}/lib/R"
+
+    system "Rscript", "-e", "install.packages(c('ggplot2','doMC','data.table'),repos='https://cran.rstudio.com')"
+    system "R", "CMD", "INSTALL", "./package/DASTool_1.1.1.tar.gz"
+
+    system "unzip", "db.zip", "-d", "db"
+
+    chmod 0755, "DAS_Tool"
+    inreplace "DAS_Tool", "split --numeric-suffixes --lines=", "split -l " if OS.mac?
+
+    libexec.install "DAS_Tool", "src", "db", "lib"
+    (bin/"DAS_Tool").write_env_script libexec/"DAS_Tool", :R_LIBS_SITE => libexec/"lib/R"
+  end
+
+  test do
+    assert_match "Usage", shell_output("#{bin}/DAS_Tool -h", 1)
+  end
+end


gaberoo:
OK, I see the build error on Linux, and will look into it. Seems like an easy fix.

```
==> R CMD install ./package/DASTool_1.1.1.tar.gz
install: missing destination file operand after './package/DASTool_1.1.1.tar.gz'
Try 'install --help' for more information.
```

@sjackman: Before I push any new changes though, two questions:
1) Is this worth pursuing
2) The build stage is quite long because R packages need to be installed. Are there any CircleCI ENVs that I should set?

Formula/das_tool.rb
```diff
@@ -0,0 +1,40 @@
...
+  depends_on "diamond"
+  depends_on "prodigal"
+  depends_on "pullseq"
+  depends_on "r"
+
+  def install
+    mkdir_p buildpath/"lib/R"
+    ENV["R_LIBS_SITE"] = "#{buildpath}/lib/R"
+
+    system "Rscript", "-e", "install.packages(c('ggplot2','doMC','data.table'), repo='http://cran.us.r-project.org');"
```
sjackman: How about using `repo='https://cran.rstudio.com'` ? I don't know how it's like now, but I believe it may be faster than `cran.us.r-project.org`.

sjackman:
```
Unwanted system libraries:
  /usr/lib/x86_64-linux-gnu/libicudata.so.55
  /usr/lib/x86_64-linux-gnu/libicui18n.so.55
  /usr/lib/x86_64-linux-gnu/libicuuc.so.55
==> ‌FAILED‌
```
https://dev.azure.com/brewsci/brewsci/_build/results?buildId=372

Try adding a dependency on `icu4c`. That might help. I suspect it's the R libraries that are linking against these libraries.

sjackman:
Looks great! If you're able to sort out the i18n linkage issue, it's good to go.

sjackman:
Do you know about the `linuxbrew/brew` Docker image? It's useful for troubleshooting build failures like this if you're not able to reproduce it on your local machine. Try running `brew linkage das_tool‌` and brew linkage --reverse das_tool` on your local machine.

gaberoo:
I'll have to check out that docker image. Will come in handy.

Otherwise I think this is now good to go, @sjackman.

Formula/das_tool.rb
```diff
@@ -0,0 +1,42 @@
...
+  # cite Sieber_2018: "https://doi.org/10.1038/s41564-018-0171-1"
+  desc "Genomic binning refiner"
+  homepage "https://github.com/cmks/DAS_Tool"
+  url "https://github.com/cmks/DAS_Tool/archive/1.1.1.tar.gz"
+  sha256 "2a55f67b5331251d8fd5adea867cc341363fbf7fa7ed5c3ce9c7679d8039f03a"
+  head "https://github.com/cmks/DAS_Tool.git"
+
+  if OS.mac?
+    depends_on "coreutils"
+    depends_on "gettext"
```
sjackman: Is there run-time linkage against `gettext` on macOS? Check `brew linkage das_tool`.

Formula/das_tool.rb
```diff
@@ -0,0 +1,42 @@
+class DasTool < Formula
+  # cite Sieber_2018: "https://doi.org/10.1038/s41564-018-0171-1"
+  desc "Genomic binning refiner"
+  homepage "https://github.com/cmks/DAS_Tool"
+  url "https://github.com/cmks/DAS_Tool/archive/1.1.1.tar.gz"
+  sha256 "2a55f67b5331251d8fd5adea867cc341363fbf7fa7ed5c3ce9c7679d8039f03a"
+  head "https://github.com/cmks/DAS_Tool.git"
+
+  if OS.mac?
+    depends_on "coreutils"
```
sjackman: Could you open a bug report upstream with `DAS_Tool` that it doesn't work with the `/usr/bin/split` provided by macOS? This dependency is fine in the mean time.

Formula/das_tool.rb
```diff
@@ -0,0 +1,42 @@
...
+    depends_on "gettext"
+  else
+    depends_on "icu4c"
+    depends_on "unzip" => :build
+  end
+
+  depends_on "diamond"
+  depends_on "prodigal"
+  depends_on "pullseq"
+  depends_on "r"
```
sjackman: On macOS, it's much better to use the cask `r-app` rather than the brewed `r`, because the former can use precompiled binary packages from CRAN, but the latter cannot, due to its installation directory.
Let's make this a `:build` dependency on macOS. You could add a `def caveats` that instructs the user to `brew cask install r-app` on macOS. Not an ideal situation I know.

Formula/das_tool.rb
gaberoo: There is. Maybe this is the same dependency as `icu4c` on Linux?

```
❯ brew linkage das_tool
System libraries:
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
  /usr/lib/libSystem.B.dylib
  /usr/lib/libc++.1.dylib
Homebrew libraries:
  /usr/local/opt/gettext/lib/libintl.8.dylib (gettext)
  /usr/local/opt/r/lib/R/lib/libR.dylib (r)
```

sjackman: Thanks for checking. Both `icu4c` and `gettext` are related to internationalization, but they are different packages. macOS includes `icu4c` with the OS but not `gettext`, and Linux is the opposite situation.

Formula/das_tool.rb
sjackman: Assuming `brew test das_tool` passes without R installed. There's a workaround if it fails. You can add dependencies for testing only.

Formula/das_tool.rb
gaberoo: Done.

https://github.com/cmks/DAS_Tool/issues/36

Also, I've changed the patch in the formula to the one I proposed in the Issue. This then uses BSD-compatible split flags.

sjackman: Thanks!

Formula/das_tool.rb
gaberoo: The test is somewhat of a pseudo-test, so it should pass. I'm wondering then you if you would prefer the user to add the packages manually, rather than install them in the package directory. Will the bottle linkages work if `r-app` is installed, rather than brew `r`?

```
❯ brew linkage --reverse das_tool
...
/usr/local/opt/r/lib/R/lib/libR.dylib
  libexec/lib/R/Rcpp/libs/Rcpp.so
  libexec/lib/R/colorspace/libs/colorspace.so
  libexec/lib/R/data.table/libs/datatable.so
  libexec/lib/R/digest/libs/digest.so
  libexec/lib/R/fansi/libs/fansi.so
  libexec/lib/R/glue/libs/glue.so
  libexec/lib/R/lazyeval/libs/lazyeval.so
  libexec/lib/R/plyr/libs/plyr.so
  libexec/lib/R/reshape2/libs/reshape2.so
  libexec/lib/R/rlang/libs/rlang.so
  libexec/lib/R/scales/libs/scales.so
  libexec/lib/R/stringi/libs/stringi.so
  libexec/lib/R/tibble/libs/tibble.so
  libexec/lib/R/utf8/libs/utf8.so
```

gaberoo: I pushed the change, though I'm unsure if it'll work. Perhaps best to write a proper test. There's example data, though it's not a simple test.

sjackman: > I'm wondering then you if you would prefer the user to add the packages manually, rather than install them in the package directory. Will the bottle linkages work if r-app is installed, rather than brew r?

Hmm. Right. If there were no native (compiled) packages, it'd work just fine. Since there are native packages, you're right, it probably won't work as is.

The two options are to use the brewed R on both Linux and macOS and install the packages as you've done on both. Alternatively, we could install the packages only for Linux, and on macOS display a caveat instructing the user to `brew cask install r-app` and `Rscript -e "install.packages(…)"`. Which do you prefer?

sjackman: I have a slight preference for the latter. I generally think it's a mistake to install the brewed R on macOS, since the cask works better, and having both installed causes confusion.

gaberoo: I'm fine with either. Personally I've been using the brewed R for a while, so I would have leaned towards that. Maybe the very best would be to put a runtime check for the packages in a wrapper script that notifies the user if they are not installed?

gaberoo: Something like this?

```
!/bin/bash
Rscript -e 'library(ggplot2);library(data.table);library(doMC);' 2&>1 /dev/null
if [ $? -ne 0 ]; then
  echo "DAS_Tool requires the packages 'ggplot2', 'data.table', "
  echo "and 'doMC' to be installed. Please install them before use "
  echo "with "
  echo ""
  echo "  Rscript -e 'install.packages(c(\"ggplot2\",\"data.table\",\"doMC\"))'"
  echo ""
else
```

gaberoo:
This would have been a possible test, but it takes too long to complete.

```
  def install
    ...
    pkgshare.install "sample_data"
  end

  test do
    assert_match "Usage", shell_output("#{bin}/DAS_Tool -h", 1)

    cp_r pkgshare/"sample_data", "."
    system "#{bin}/DAS_Tool", "-i", "sample_data/sample.human.gut_concoct_scaffolds2bin.tsv,sample_data/sample.human.gut_maxbin2_scaffolds2bin.tsv",
                              "-c", "sample_data/sample.human.gut_contigs.fa",
                              "-o", "test"
    assert_match "bin selection complete: 9 bins above score threshold selected.", "test_DASTool.log"
  end
```

Formula/das_tool.rb
sjackman: Sure that works for me. For style points, I'd suggest using `cat <<EOS` rather than `echo`.

gaberoo: Actually, I just realised that the package won't install unless R is already installed:

```
 system "R", "CMD", "INSTALL", "./package/DASTool_1.1.1.tar.gz"
```

Is there a way to require either 'r-app' or 'r' be installed a priori?

sjackman: Let's use the brewed R on both macOS and Linux. Sounds simpler. Sorry for the runaround.

sjackman:
Merged. Thanks again, Gabriel! 🏆

PR #573, opened 2019-03-16
staden-io-lib 1.14.11 (new formula)

gaberoo:
- [ x ] Does your build pass `brew audit --strict FORMULA` (after doing `brew install FORMULA`)?

-----

This is a dependency of salmon.

### DIFF
--- a/Formula/staden-io-lib.rb
+++ b/Formula/staden-io-lib.rb
@@ -1,0 +1,44 @@
+class StadenIoLib < Formula
+  desc "Staden Package io_lib"
+  homepage "https://staden.sourceforge.io/"
+  url "https://github.com/jkbonfield/io_lib/releases/download/io_lib-1-14-11/io_lib-1.14.11.tar.gz"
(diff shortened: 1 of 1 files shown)


gaberoo:
Interesting: The test returns exit code 0 on macOS and 1 on Linux.

sjackman:
> This is a dependency of salmon.

No kidding. @rob-p Do you prefer to htslib?

sjackman:
The name `io_lib` is too generic. Debian and variants calls this library `staden-io-lib`. Let's use that name. See https://repology.org/project/staden-io-lib/versions

rob-p:
I do. The reasons are twofold. First, io_lib supported multithreaded bam decompression long before htslib (though mut point now). Second, however, it adopts a different encoding of bam records that is more memory efficient than that of htslib.

sjackman:
`brew test` failed. See the Azure build logs.

gaberoo:
@sjackman: quite strange about the test. In any case, I wrote a much more minimal one.

@rob-p: Interesting about the encoding of BAM records, can you elaborate?

sjackman:
Merged. Thanks again, Gabriel!

PR #475, opened 2018-10-09
readseq 2.1.30_1

tseemann:
I want this bottled as the .jar file often fails to download from the home page.

PR #596, opened 2019-04-29
readseq: Remove this formula

sjackman:
The web site hosting the homepage and JAR file are not responding.
See also PR https://github.com/brewsci/homebrew-bio/pull/475

$ brew fetch readseq
curl: (7) Failed to connect to iubio.bio.indiana.edu port 80: Operation timed out


tseemann:
Yes, it is unfortunate, but it has to go. I can't find another location for the JAR file.

PR #599, opened 2019-05-02
crass 1.0.1

gaberoo:

-----

Question for @sjackman:
What is the policy about putting things in `pkgshare` vs `share` ?

### DIFF
--- a/Formula/crass.rb
+++ b/Formula/crass.rb
@@ -1,0 +1,34 @@
+class Crass < Formula
+  # cite Skennerton_2013: "https://doi.org/10.1093/nar/gkt183"
+  desc "The CRISPR assembler"
+  homepage "https://ctskennerton.github.io/crass"
+  url "https://github.com/ctSkennerton/crass/archive/v1.0.1.tar.gz"
+  sha256 "7078bcccb7a16bf4f2e54590f25d85f36d27d1bbaa9f2c484d4eaa81a5ead930"
+  head "https://github.com/ctSkennerton/crass.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+
+  depends_on "graphviz"
+  depends_on "xerces-c"
+  depends_on "zlib" unless OS.mac?
+
+  def install
+    system "./autogen.sh"
+    system "./configure", "--enable-rendering", "--prefix=#{prefix}"
+
+    system "make", "install"
+
+    pkgshare.install "test"
+    pkgshare.install "scripts"
+    pkgshare.install "doc/manual.tex"
+  end
+
+  test do
+    assert_match "Usage", shell_output("#{bin}/crass", 1)
+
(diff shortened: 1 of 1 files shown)


sjackman:
> What is the policy about putting things in pkgshare vs share ?

Always use `pkgshare`. The only two directories under `share` should be the package name and `man`.

sjackman:
```
/home/linuxbrew/.linuxbrew/Cellar/fontconfig/2.13.1/bin/fc-cache: failed
Warning:‌ The post-install step did not complete successfully‌
You can try again using `brew postinstall fontconfig`
```
https://dev.azure.com/brewsci/brewsci/_build/results?buildId=503&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=8ef82b3b-1feb-5bbd-06f6-b1f7b5467f03&lineStart=4350&lineEnd=4351&colStart=1&colEnd=1

sjackman:
CircleCI on Linux
```
checking for Xerces C++ Parser... no
configure: error: Cannot find Xerces-c
```
https://circleci.com/gh/brewsci/homebrew-bio/5768

I see that you do have `depends_on "xerces-c"` but it still doesn't seem to be finding it. You can troubleshoot using the `linuxbrew/brew` Docker image, used by CircleCI.

gaberoo:
I'll check the `fontconfig` error. But the more worrying one is

```
checking for Xerces C++ Parser... no
configure: error: Cannot find Xerces-c
```

on both Azure and CircleCI Linux builds.

gaberoo:
There are two problems now:

1. `xerces-c` is still not found on Linux. This must be a problem with the `m4` file or the way `xerces-c` is install on Linux versus macOS. One thing I did observe is that on linux `libxerces-c.a` is not symlinked, only the versioned library `libxerces-c-3.2.a`.

2. The newest `netpbm` does not build on Linux.

sjackman:
> The newest netpbm does not build on Linux.

I've asked Michka to take a look at `netpbm`. https://github.com/Homebrew/linuxbrew-core/pull/13078#issuecomment-496702619

sjackman:
On macOS `xerces-c` installs…
```
lib/libxerces-c-3.2.dylib
lib/libxerces-c.so
lib/pkgconfig/xerces-c.pc
```
Pretty weird that it installs an unversioned `.so` file on macOS rather than a `.dylib` file.

My guess is that `pkg-config` should be used to find the correct compiler flags. Is that working as expected on Linux? Is `crass` using `pkg-config` to find `xerces-c`?

gaberoo:
`pkgconfig/xerces-c.pc` is not installed on Linux.

sjackman:
> `lib/pkgconfig/xerces-c.pc` is not installed on Linux.

If `crass` does use `pkg-config` to find `xerces-c`, then I feel like that not installing the `.pc` file on Linux may the root of the problem.

PR #613, opened 2019-05-09
metabat 2.13 (new formula)

gaberoo:

-----

Some comments:

- The formula should build hassle-free on Linux
- On macOS, there are some issues:
    1. The current release requires _Atomic types (`u_int`) etc. which are broken in Mojave. This is why I'm using the current master head.
    2. `FindBoost.cmake` is doing weird things and selecting boost from `/usr/local/` despite explicitly being told not to do so.

The result is somewhat ugly, but it builds and runs on macOS. Interestingly, the `llvm_clang` + `libomp` build segfaulted on real data.

@sjackman: What do you think?

### DIFF
--- a/Formula/metabat.rb
+++ b/Formula/metabat.rb
@@ -1,0 +1,80 @@
+class Metabat < Formula
+  # cite Kang_2015: "https://doi.org/10.7717/peerj.1165"
+  desc "Statistical framework for reconstructing genomes from metagenomic data"
+  homepage "https://bitbucket.org/berkeleylab/metabat/"
+  url "https://bitbucket.org/berkeleylab/metabat/get/v2.13.tar.gz"
+  sha256 "aa75a2b62ec9588add4c288993821bab5312a83b1259ff0d508c215133492d74"
+  head "https://bitbucket.org/berkeleylab/metabat.git"
+
+  depends_on :linux unless build.head?
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+
+  fails_with :clang # needs OpenMP
+
+  if OS.mac?
+    resource "boost" do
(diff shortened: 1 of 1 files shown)


Formula/metabat.rb
```diff
@@ -0,0 +1,69 @@
...
+  version "2.13-b3bd9c2"
+  sha256 "4b3e59c28875ee0002164705a1c91d8643b904d0ee7df16e3c92ef9c55b6b16c"
+  head "https://bitbucket.org/berkeleylab/metabat.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+
+  if OS.mac?
+    fails_with :clang
```
sjackman: This failure should probably be outside the `if OS.mac?` block. If it's because it requires OpenMP, I like adding a comment.
```ruby
    fails_with :clang # needs openmp
```

sjackman:
I suggest using the most recent stable release for the `stable` spec and the current master head to `head`. Add `depends_on :linux unless build.HEAD?`

gaberoo:
OK, it's not `HEAD` only on macOS. I added mac-specific caveats to indicate this.

Formula/metabat.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  depends_on :linux unless build.head?
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+
+  fails_with :clang # needs OpenMP
+
+  if OS.mac?
```
sjackman: Please add a one-line comment here explaining that `boost` needs to be compiled with GCC.

sjackman: Good to merge after that. Thank you, Gabriel!

PR #651, opened 2019-05-27
das_tool: fix env path

gaberoo:
A script file `Fasta_to_Scaffolds2Bin.sh` was using `/bin/env bash` in the hash bang. I missed this last time. I've also added an `ln_s` for this script as it's referenced in https://github.com/cmks/DAS_Tool/blob/master/README.md.

Is a revision bump needed for such fixes?

### DIFF
--- a/Formula/das_tool.rb
+++ b/Formula/das_tool.rb
@@ -4,6 +4,7 @@ class DasTool < Formula
   homepage "https://github.com/cmks/DAS_Tool"
   url "https://github.com/cmks/DAS_Tool/archive/1.1.1.tar.gz"
   sha256 "2a55f67b5331251d8fd5adea867cc341363fbf7fa7ed5c3ce9c7679d8039f03a"
+  revision 1
   head "https://github.com/cmks/DAS_Tool.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


jonchang:
> Is a revision bump needed for such fixes?

Any time users need to get new bottles via `brew upgrade` the revision should be bumped.

Formula/das_tool.rb
```diff
@@ -36,12 +37,15 @@ def install
 
     chmod 0755, "DAS_Tool"
     inreplace "DAS_Tool", "split --numeric-suffixes --lines=", "split -l " if OS.mac?
+    inreplace "src/Fasta_to_Scaffolds2Bin.sh", "#!/bin/env bash", "#!/bin/bash"
```
sjackman: This PR is fine. I'd suggest opening an upstream PR to change `#!/bin/env bash` to either `#!/usr/bin/env bash` which is more portable or `#!/bin/bash` which is simpler and good enough.

gaberoo: https://github.com/cmks/DAS_Tool/pull/40

PR #678, opened 2019-07-11
bowtie 1.2.3

dmohadjel:

-----

`brew audit --strict bowtie` will fail. Two particular issues:
```
Stable: version 1.2.3 is redundant with version scanned from URL
```
I was able to resolve this one by removing the `version` flag in the file. 
The other problem:
```
* Non-executables were installed to "/home/dmohadjel/.linuxbrew/opt/bowtie/bin"
    The offending files are:
      /home/dmohadjel/.linuxbrew/opt/bowtie/bin/bowtie-buildc
```

### DIFF
--- a/Formula/bowtie.rb
+++ b/Formula/bowtie.rb
@@ -14,13 +14,13 @@ class Bowtie < Formula
   end
 
   depends_on "tbb"
+  depends_on "python@2" => :test unless OS.mac? || which("python")
 
   def install
     system "make", "install", "prefix=#{prefix}"
 
     doc.install "MANUAL", "NEWS", "TUTORIAL"
     pkgshare.install "scripts", "genomes", "indexes", "reads"
-    bin.install "bowtie-inspect"
   end
 
   test do
--- a/Formula/bowtie.rb
+++ b/Formula/bowtie.rb
@@ -20,6 +20,7 @@ class Bowtie < Formula
 
     doc.install "MANUAL", "NEWS", "TUTORIAL"
     pkgshare.install "scripts", "genomes", "indexes", "reads"
+    bin.install "bowtie-inspect"
   end
 
   test do
--- a/Formula/bowtie.rb
+++ b/Formula/bowtie.rb
@@ -2,9 +2,8 @@ class Bowtie < Formula
   # cite Langmead_2009: "https://doi.org/10.1186/gb-2009-10-3-r25"
   desc "Ultrafast memory-efficient short read aligner"
   homepage "https://bowtie-bio.sourceforge.io/"
-  url "https://github.com/BenLangmead/bowtie/archive/v1.2.2_p1.tar.gz"
-  version "1.2.2_p1"
-  sha256 "e1b02b2e77a0d44a3dd411209fa1f44f0c4ee304ef5cc83f098275085740d5a1"
+  url "https://github.com/BenLangmead/bowtie/archive/v1.2.3.tar.gz"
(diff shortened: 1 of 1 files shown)


dmohadjel:
CircleCI for Linux

```
==> brew test brewsci/bio/bowtie --verbose
==> FAILED
Testing brewsci/bio/bowtie
==> /home/linuxbrew/.linuxbrew/Cellar/bowtie/1.2.3/bin/bowtie-inspect --help 2>&1
Error: brewsci/bio/bowtie: failed
An exception occurred within a child process:
  Test::Unit::AssertionFailedError: <0> expected but was
<127>.
```

sjackman:
What type of file is `bowtie-buildc`? It's probably missing the shebang line. If it's a shell script, add `#!/bin/sh` to the start of the file using `inreplace`.

sjackman:
`bowtie-inspect` is a Python script. I believe that exit status `<127>` means that the file is missing. That's okay if it's meant to be missing, if it was intentionally removed in this version. Is it anywhere in the source tarball?

dmohadjel:
> What type of file is `bowtie-buildc`? It's probably missing the shebang line. If it's a shell script, add `#!/bin/sh` to the start of the file using `inreplace`.

It seems to be a `.pyc` file. Using `file bowtie-buildc` returns `python 2.7 byte-compiled`.

> `bowtie-inspect` is a Python script. I believe that exit status `<127>` means that the file is missing. That's okay if it's meant to be missing, if it was intentionally removed in this version. Is it anywhere in the source tarball?

There is a Python script text executable named `bowtie-inspect` in the source tarball.

sjackman:
> It seems to be a .pyc file. Using file bowtie-buildc returns python 2.7 byte-compiled.

My best guess is that it's a bug in their build script. They're renaming `*.py` files to `*`, because you should remove the `.py` extension of Python scripts when they're installed. The unintentionally renaming `bowtie-build.pyc` to `bowtie-buildc` by deleting the `.py`. That's just my best guess. If there's also a Python script installed named `bowtie-build`, you can delete the file named `bowtie-buildc`.

sjackman:
> There is a Python script text executable named bowtie-inspect in the source tarball.

You can install it with `bin.install "bowtie-inspect"`.

dmohadjel:
CircleCI for Linux ran into the same issue. When I tried `brew test bowtie` locally it passed the first couple of times, but now when I tested it again I received the same error, except with exit status `<126>` rather than `<127>`. I'm not too familiar with Ruby, am I implementing `bin.install "bowtie-inspect"` correctly here? 

> My best guess is that it's a bug in their build script. They're renaming `*.py` files to `*`, because you should remove the `.py` extension of Python scripts when they're installed. The unintentionally renaming `bowtie-build.pyc` to `bowtie-buildc` by deleting the `.py`. That's just my best guess. If there's also a Python script installed named `bowtie-build`, you can delete the file named `bowtie-buildc`.

I deleted `bowtie-buildc` and the audit passed.

sjackman:
Ah, the problem is probably not that `bowtie-inspect` is not installed. The problem is probably that `python` is not installed.
```
❯❯❯ head -n1 /usr/local/bin/bowtie-inspect
#!/usr/bin/env python
```
Normally we'd add a dependency on `python@2`. I don't really want to add a dependency on `python@2` just for this one script `bowtie-inspect`, that many people probably won't use. I suggest…
```ruby
depends_on "python@2" => :test unless OS.mac? || which("python")
```
You can remove `bin.install "bowtie-inspect"`

sjackman:
Merged. Thanks again, @dmohadjel ! 🏆

PR #836, opened 2019-11-18
rasusa 0.1.0 (new formula)

mbhall88:

-----

The only audit check it doesn't pass is `Stable: version 0.1.0 is redundant with version scanned from URL` as I have manually specified the version number.

Closes https://github.com/mbhall88/rasusa/issues/13

### DIFF
--- a/Formula/rasusa.rb
+++ b/Formula/rasusa.rb
@@ -12,8 +12,6 @@ class Rasusa < Formula
     sha256 "eb6d1c1ab48bd3fd568fd7dad338352c54a514762502fc2ec4172ae792a77832"
   end
 
-  conflicts_with "rasusa"
-
   def install
     bin.install "rasusa"
   end
--- a/Formula/rasusa.rb
+++ b/Formula/rasusa.rb
@@ -2,13 +2,12 @@ class Rasusa < Formula
   # cite Hall_2019: "https://doi.org/10.5281/zenodo.3546168"
   desc "Randomly subsample sequencing reads to a specified coverage"
   homepage "https://github.com/mbhall88/rasusa"
-  version "0.1.0"
 
   if OS.mac?
-    url "https://github.com/mbhall88/rasusa/releases/download/#{version}/rasusa-#{version}-x86_64-apple-darwin.tar.gz"
+    url "https://github.com/mbhall88/rasusa/releases/download/0.1.0/rasusa-0.1.0-x86_64-apple-darwin.tar.gz"
     sha256 "f485fe769c6170020bd7dc65205a7139dcecff03f4808a9241bd2843a36b6e83"
   elsif OS.linux?
-    url "https://github.com/mbhall88/rasusa/releases/download/#{version}/rasusa-#{version}-x86_64-unknown-linux-musl.tar.gz"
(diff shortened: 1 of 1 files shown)


mbhall88:
The checks seem to fail now purely based on the audit check mentioned above about the version number. If I remove the version specification it doesn't parse out the correct version, so it has to be specified.

Formula/rasusa.rb
```diff
@@ -0,0 +1,21 @@
...
+    url "https://github.com/mbhall88/rasusa/releases/download/0.1.0/rasusa-0.1.0-x86_64-unknown-linux-musl.tar.gz"
+    sha256 "eb6d1c1ab48bd3fd568fd7dad338352c54a514762502fc2ec4172ae792a77832"
+  end
+
+  def install
+    bin.install "rasusa"
+  end
+
+  test do
+    system "#{bin}/rasusa", "--help"
```
sjackman:
```suggestion
    assert_match "Usage", shell_output("#{bin}/rasusa --help")
```

sjackman:
> The checks seem to fail now purely based on the audit check mentioned above about the version number. If I remove the version specification it doesn't parse out the correct version, so it has to be specified.

That's quite strange. The version number `0.1.0` seems correct on CircleCI.
https://app.circleci.com/jobs/github/brewsci/homebrew-bio/7946

sjackman:
I'd much prefer to build this program from source using the brewed `rust` as a build dependency.
```ruby
depends_on "rust" => :build
```

mbhall88:
> I'd much prefer to build this program from source using the brewed `rust` as a build dependency.
> 
> ```ruby
> depends_on "rust" => :build
> ```

 My reasoning for doing it this way though was that
1. The user doesn't need to have `rust` installed this way
2. This is the same brew strategy used by [`ripgrep`](https://github.com/BurntSushi/ripgrep/blob/master/pkg/brew/ripgrep-bin.rb) which is a very heavily used rust tool.

I can change it to build from source with `cargo` if you like though.

mbhall88:
@sjackman so this won't be merged unless I build from source?

tseemann:
@mbhall88 I think @sjackman is just hoping we can start to migrate `rust` and `go` tools to be built from source now. it wasn't possible in the past as the compilers weren't properly within brew. the user doesn't need the compilers installed on their desktop, because the will still install binaries, but the CI/TRAVIS build system will compile from source rather than download Brent's binaries. 

I'm not against merging this as-is.  @sjackman ? @jonchang ?

jonchang:
Think it should build from source rather than install from binaries. I’m with @sjackman on this one.

Bottles shouldn’t need rust as a run time dependency.

mbhall88:
Ok, no worries. I have no idea how to make brew recipes (I effectively copied this from another rust tool) so I will find another one that builds from source and update this PR.

mbhall88:
Right, I _think_ this recipe should do it. Please let me know if there is anything else I should be doing in the recipe.

tseemann:
@mbhall88 don't worry, we all just copy other recipes - it's the only sane way to do it :)
And the build from source recipe is shorter, and will be easier to update version too.
Nice work!

tseemann:
@mbhall88 thank you - it is now part of brew.

PR #851, opened 2019-12-20
circle: update to latest Xcode on Mojave

jonchang:
c.f. https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

PR #856, opened 2020-01-07
Revert "circle: update to latest Xcode on Mojave"

sjackman:
Reverts brewsci/homebrew-bio#851

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           path: /tmp/bottles
   build-macos:
     macos:
-      xcode: "11.1.0"
+      xcode: "10.2.0"
     environment:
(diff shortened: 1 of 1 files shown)


sjackman:
Torsten opened PR https://github.com/brewsci/homebrew-bio/pull/799 to updated to updated XCode to version 10.3.0, which also unintentionally updated from Sierra to Mojave. At my request, he opened PR https://github.com/brewsci/homebrew-bio/pull/800 to change XCode version from 10.3.0 to 10.2.0 to continue running on Sierra, but as it turns out it was actually running on Mojave. So as a result we've been building bottles for Mojave since Oct 16. That being the case, let's stick with Mojave.

Closing without merging.

PR #872, opened 2020-01-23
Add the required dependencies python (3) and perl (viennarna)

hwalinga:

-----

viennarna will not install unless you add the dependencies "python" (version 3) and "perl". Without "python" it will complain it cannot find "Python.h", and without "perl" it will complain it cannot find "EXTERN.h". 

It might also be a good idea to point this to the newest version (2.4.3 -> 2.4.14), but maybe you want to have this in a different PR.


hwalinga:
The build failed because it complains that python and python@2 are conflicting dependencies. However, they are not conflicting and can just be installed both. I also tested removing python@2 from the dependencies, but they are really required both.

sjackman:
Does CI fail if you remove `python@2` from the dependencies? Can you try removing it again, so that I can see the CI failure?

hwalinga:
Okay, I don't see how it helps, but you might know what you want with it.

 I can only tell you that if I try to install it without the python@2 dependency it will fail on my PC. The package will build interfaces for both Perl, Python2, and Python3. 

It's also interesting to note that if you remove python@2 it will fail in the "Making all in Python" step and if you remove python it will fail in the "Making all in Python", while it both searches for Python.h. I don't how everything works with linking and so on, but apparently it works.

hwalinga:
Interesting that only macOS failed but not Linux: 

```
Could not link test program to Python2. Maybe the main Python2 library has been
```

sjackman:
```
checking consistency of all components of python development environment... no
configure: error: in `/private/tmp/viennarna-20200125-5222-1qmq2v6/ViennaRNA-2.4.3':
configure: error: 
  Could not link test program to Python2. Maybe the main Python2 library has been
  installed in some non-standard library path. If so, pass it to configure,
  via the LIBS environment variable.
  Example: ./configure LIBS="-L/usr/non-standard-path/python/lib"
  ============================================================================
   ERROR!
   You probably have to install the development version of the Python package
   for your distribution.  The exact name of this package varies among them.
  ============================================================================

See `config.log' for more details
```
https://github.com/brewsci/homebrew-bio/pull/872/checks?check_run_id=408593322#step:3:814

Is there a `configure` option to disable Python 2? See `./configure --help`
Are these Perl and Python plugins necessary? Perhaps the Perl and Python 3 plugins should also be disabled.

hwalinga:
Yes, I looked it up, and there is: --without-perl, --without-python, --without-python3. 

I am just confused as it just works on my pc with the depends_on lines. 

I personally only need the binaries. Don't know about other users.

sjackman:
Please add `./configure --without-perl --without-python --without-python3` and remove these dependencies.

hwalinga:
Let's do a compromise. Python2 has recently been deprecated, so what if we only remove that dependency. Maybe it will not complain with only a python in de dependencies.

hwalinga:
The issue is more specific this time: 

```
Error: 1 problem in 1 formula detected
brewsci/bio/viennarna:
  * python modules have explicit framework links
    These python extension modules were linked directly to a Python
    framework binary. They should be linked with -undefined dynamic_lookup
    instead of -lpython or -framework Python.
      /usr/local/opt/viennarna/lib/python3.7/site-packages/_RNA.cpython-37m-darwin.so
==> FAILED
```

Do you know how such a linking should be defined?

sjackman:
It sounds like somewhere in the upstream ViennaRNA build system there's either `-lpython` or `-framework Python` which should be replaced with `-undefined dynamic_lookup`. If you're able to find the offending line you can fix it with either `inreplace "filename", "old text", "new text"` or a `patch do` block. Alternatively you could add `--without-python3`.

hwalinga:
The new version (2.4.14) uses `-undefined dynamic_lookup` instead of `-lpython`, let's bump the version instead of fixing the old one. 

I don't know what the exact steps for a version bump would be. I assume the bottle should now be updated as well?

hwalinga:
It worked.

My only concern now is that the `depends_on "python"` and the `"--without-python"` seem contradicting. Maybe we should add a comment indicating that it depends on PY3 and is build without PY2.

sjackman:
Merged! Thank you for your contribution to Brewsci/bio, Hielke! 🌟

hwalinga:
Great, thanks for thinking with me about this one. 

But you are sure you don't want to add a comment to emphasize that the --without-python only refers to python2?

PR #937, opened 2020-03-03
bali-phy 3.5.0

bredelings:

-----


bredelings:
BTW, this formula requires c++17.  I presume it won't build on OS X < 10.14 unless something is done.  Perhaps I should switch to GCC.

bredelings:
On linux, it seems like compiling with gcc-9 is problematic, since the dynamic libs are bound to the system libstdc++.so.6 by default.  However, the default of gcc-5 is way too old to handle c++17.

PR #1000, opened 2020-04-22
xmatchview 1.2.3

lcoombe:

-----

### DIFF
--- a/Formula/xmatchview.rb
+++ b/Formula/xmatchview.rb
@@ -2,8 +2,8 @@ class Xmatchview < Formula
   # cite L_Warren_2018: "https://doi.org/10.21105/joss.00497"
   desc "Smith-waterman alignment visualization"
   homepage "https://github.com/bcgsc/xmatchview"
-  url "https://github.com/bcgsc/xmatchview/archive/v1.2.0.tar.gz"
-  sha256 "17723ba8d6162c2b57f9c63f1f93fcc878963eb00097a12a575c8b3b3c9ef1bf"
+  url "https://github.com/bcgsc/xmatchview/archive/v1.2.3.tar.gz"
+  sha256 "774dd0f07946511b853eed167d3fbff56f72fa0ee0f4a586207904246789042e"
   head "https://github.com/bcgsc/xmatchview.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


lcoombe:
macOS error:
```
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/tmp/pip-install-wr6_3mwi/pillow/setup.py", line 914, in <module>
        raise RequiredDependencyException(msg)
    __main__.RequiredDependencyException:
    
    The headers or library files could not be found for zlib,
    a required dependency when compiling Pillow from source.
```

lcoombe:
I'm struggling to fix this error...It is confusing because I can't reproduce this on my local mac homebrew installation and nothing I've tried so far has helped (adding dependency on pkg-config, installing particular version, having full dependency on zlib, etc)

sjackman:
Congratulations on PR number 1000, Lauren! 🍾

sjackman:
Fortunately / unfortuantely I can reproduce this error locally on macOS 10.14 (Mojave). I'll troubleshoot and get back to you.

Do you have `/usr/include` on your macOS? I do not. That's likely the cause.

lcoombe:
Thank you very much for the troubleshooting help, Shaun!

Yes, I do have `/usr/include/` on my macOS

sjackman:
Here's the upstream bug and fix: https://github.com/python-pillow/Pillow/issues/3438#issuecomment-543812237

sjackman:
I'm not sure why Pillow is being built from source though. It looks like there is a binary package (wheel) available, `Pillow-7.1.1-cp37-cp37m-macosx_10_10_x86_64.whl ` and `Pillow-7.1.1-cp38-cp38-macosx_10_10_x86_64.whl` at https://pypi.org/project/Pillow/#files

lcoombe:
I think I built from source to attempt to solve issues I was having when I originally added `xmatchview`: https://github.com/brewsci/homebrew-bio/pull/586#issuecomment-481038979

I tried removing the build from source at some point in this troubleshooting process, but I saw different errors cropping up..

sjackman:
Success!

lcoombe:
Yay! Thanks for your help Shaun :)

PR #1082, opened 2020-07-17
shovill: fix dependencies (# 1082)

SeekingMeaning:

-----

`flash` dependency needs to be updated to `brewsci/bio/flash` since a different formulae with the same name was added to `homebrew/core`

### DIFF
--- a/Formula/shovill.rb
+++ b/Formula/shovill.rb
@@ -12,20 +12,20 @@ class Shovill < Formula
     sha256 "4c8964af516ab407aa754466600409e9e0dd262ed1804211729f34b20cfdc1ea" => :x86_64_linux
   end
 
+  depends_on "brewsci/bio/flash"
+  depends_on "brewsci/bio/kmc"
+  depends_on "brewsci/bio/lighter"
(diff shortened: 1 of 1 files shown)


sjackman:
> flash dependency needs to be updated to brewsci/bio/flash since a different formulae with the same name was added to homebrew/core

😫 I'm quite surprised that's this time is the first that this issue has come up.

sjackman:
> There should be no need to bump the revision, since nothing material has changed.

Or more accurately, there is no need to update the bottles, since nothing material has changed.
If you do update the bottles, I always recommend bumping `revision` rather than `rebuild`. I avoid using `rebuild`.

SeekingMeaning:
Oh okay so this would be a simple push to `develop`?

sjackman:
> Oh okay so this would be a simple push to develop?

Good question. I don't have a workflow yet for merging a PR without updating the bottles. I suggest mashing the squash and merge button, and modifying the log message somehow to prevent the bottle uploader action from running. Adding a space between `# 1082` would do it.

PR #1107, opened 2020-07-18
parsnp 1.5.2

SeekingMeaning:


sjackman:
```
==> ./configure --prefix=/tmp/parsnp-20200718-14335-26m7sz/parsnp-1.5.2/muscle
Failed to execute: ./configure
```
Looks like `muscle` is no longer included with `parsnp`. You can probably remove that stanza from `install`.

Formula/parsnp.rb
```diff
@@ -41,6 +39,7 @@ def install
     # see: https://github.com/brewsci/homebrew-bio/pull/362
     cd "muscle" do
       ENV.deparallelize
+      system "./autogen.sh"
```
sjackman, line 42: Is this needed? What's the error without?

sjackman: Oh, I take it that it still includes `muscle`, but `muscle` no longer includes `./configure`?
Are you interested in trying to use `libmuscle` instead of the vendored dep?

SeekingMeaning: > Oh, I take it that it still includes `muscle`, but `muscle` no longer includes `./configure`?

Yeah

> Are you interested in trying to use `libmuscle` instead of the vendored dep?

I'm not sure; the comment right above this diff is cut off:

```rb
    # we still build this, but runtime will link against libmuscle
    # see: https://github.com/brewsci/homebrew-bio/pull/362
```

sjackman: Up to you. It's probably done this way just because that was easiest and didn't require modifying the build scripts.

sjackman:
On macOS
```
==> brew audit brewsci/bio/parsnp --online --git --skip-style
90
==> FAILED
91
Error: 1 problem in 1 formula detected
92
Missing libraries:
93
  unexpected (/private/tmp/parsnp-20200724-16576-4ihhio/parsnp-1.5.2/muscle/lib/libMUSCLE-3.7.1.dylib)
94
Unexpected missing library linkage detected
95
brewsci/bio/parsnp:
96
  * parsnp has broken dynamic library links:
97
```
https://github.com/brewsci/homebrew-bio/pull/1107/checks?check_run_id=908471970#step:4:93

I just discovered that `parsnp` is the only formula that depends on `libmuscle`, and `libmuscle` is actually built from `parsnp`. I'd suggest removing the formula `libmuscle` and `parsnp` can install this library. I'm also fine with keeping both formulae, as it is now. One formula is easier to build though. If you keep both, `libmuscle` and `parsnp` need to updated together in a single PR.

SeekingMeaning:
https://github.com/brewsci/homebrew-bio/pull/1107/checks?check_run_id=910151072#step:5:87
https://github.com/brewsci/homebrew-bio/pull/1107/checks?check_run_id=910151082#step:4:88

```
==> brew audit brewsci/bio/libmuscle --online --git --skip-style
==> FAILED
Error: 1 problem in 1 formula detected
brewsci/bio/libmuscle:
  * stable sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed.
```

PR #1139, opened 2020-08-14
workflows: add update-develop

SeekingMeaning:
Updates `develop` branch when `master` is pushed to

No-op if `develop` is already updated (e.g. pushing to `master` after uploading bottles for `develop`)

### DIFF
--- a/.github/workflows/update-develop.yml
+++ b/.github/workflows/update-develop.yml
@@ -11,4 +11,5 @@ jobs:
       - uses: actions/checkout@main
         with:
           fetch-depth: 100
-      - run: git push origin origin/master:develop
+      - name: Update the develop branch
+        run: git push origin origin/master:develop
--- a/.github/workflows/update-develop.yml
+++ b/.github/workflows/update-develop.yml
@@ -1,0 +1,38 @@
+name: Update develop
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  update-develop:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 100
+      - name: Update develop
+        run: |
+          set -eu
+          sha1='${{github.event.head_commit.id}}'
+          echo sha1="$sha1"
+          echo '${{github.event.head_commit.message}}'
+          git config --global user.name "LinuxbrewTestBot"
+          git config --global user.email "testbot@linuxbrew.sh"
+          cd "$GITHUB_WORKSPACE"
+          short_sha="$(echo "$sha1" | cut -c 1-7)"
+          git branch -f "develop-$short_sha"
+          git fetch origin develop
+          git checkout develop
+          git reset --hard origin/develop
+          git merge -m "Merge master" "develop-$short_sha"
+          git status
+          if ! git push "https://LinuxbrewTestBot:${{secrets.HOMEBREW_GITHUB_API_TOKEN}}@github.com/${{github.repository}}" develop:master develop; then
+            git push --force origin "develop-$short_sha"
+            mkdir ~/bin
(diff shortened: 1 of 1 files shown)


.github/workflows/update-develop.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  update-develop:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 100
+      - name: Update develop
+        run: |
+          set -eu
```
sjackman, line 16: I recall reading that the default on GitHub Actions is `set -e -o pipefail`. So adding `set -u` could be useful. I'm not sure that it's worth the visualize noise of adding `set -u` to every `run` block, though it could be useful during development. I'm thinking more broadly here than just this one particular action.

sjackman:
Merges to `master` ought to be relatively rare. I'm not sure that we need all the same fancy logic that we have for merging and what not for merging to `develop`, which is much more common. I'd be inclined to replace all of the body action with a one-liner `git push origin origin/master:develop`.

SeekingMeaning:
... That was satisfying

.github/workflows/update-develop.yml
```diff
@@ -8,31 +8,7 @@ jobs:
...
-          if ! git push "https://LinuxbrewTestBot:${{secrets.HOMEBREW_GITHUB_API_TOKEN}}@github.com/${{github.repository}}" develop:master develop; then
-            git push --force origin "develop-$short_sha"
-            mkdir ~/bin
-            curl -fsL https://github.com/github/hub/releases/download/v2.14.1/hub-linux-amd64-2.14.1.tgz \
-              | tar xzO hub-linux-amd64-2.14.1/bin/hub >~/bin/hub
-            chmod +x ~/bin/hub
-            git reset --hard origin/develop
-            GITHUB_TOKEN='${{secrets.HOMEBREW_GITHUB_API_TOKEN}}' ~/bin/hub pull-request -b develop -h "develop-$short_sha" -m "Update develop to commit $short_sha" --no-edit
-          fi
+      - run: git push origin origin/master:develop
```
sjackman, line 14:
```suggestion
      - name: Update develop
        run: git push origin origin/master:develop
```

SeekingMeaning:
So this merge also updated `formula.json` which was unintended but okay I think. Although, it's strange how GitHub Actions doesn't trigger CI when it pushes since pushing to `develop` didn't trigger the `upload-bottles` workflow. I'm guessing `deploy-formula-json` was originally intended to activate even with GitHub Actions pushes or perhaps I'm wrong?

sjackman:
> So this merge also updated formula.json which was unintended but okay I think.

Yes, that's fine.

> Although, it's strange how GitHub Actions doesn't trigger CI when it pushes since pushing to develop didn't trigger the upload-bottles workflow.

If you push using the GitHub App installation access token `secrets.GITHUB_TOKEN` then additional actions are _not_ triggered. If you pushing using the PAT `secrets.HOMEBREW_GITHUB_API_TOKEN` then additional actions _are_ triggered.

> I'm guessing deploy-formula-json was originally intended to activate even with GitHub Actions pushes or perhaps I'm wrong?

This PR was merged to `master`, so it triggered `deploy-formula-json`. Yes, GitHub Actions pushes are also intended to trigger `deploy-formula-json`.

SeekingMeaning:
> If you pushing using the PAT `secrets.HOMEBREW_GITHUB_API_TOKEN` then additional actions _are_ triggered.

But why doesn't `upload-bottles` trigger it? :thinking:

https://github.com/brewsci/homebrew-bio/blob/develop/.github/workflows/upload-bottles.yml#L56

PR #1203, opened 2020-11-12
upload-bottles.yml: check for `ci-skip-upload` label

SeekingMeaning:

-----

### DIFF
--- a/.github/workflows/upload-bottles.yml
+++ b/.github/workflows/upload-bottles.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   upload-bottles:
     runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, '[skip upload]')"
     container:
       image: homebrew/ubuntu16.04:master
     env:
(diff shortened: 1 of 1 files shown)


sjackman:
I'd prefer to avoid having these `[skip upload]` commands in the permanent Git log. I'm okay with running the `upload-bottles` even when it's not needed if it does no harm.

SeekingMeaning:
Perhaps a label would work?

sjackman:
A label sounds great.

.github/workflows/upload-bottles.yml
```diff
@@ -28,6 +28,11 @@ jobs:
           echo sha1='${{github.event.head_commit.id}}'
           echo '${{github.event.head_commit.message}}'
           pr="$(echo '${{github.event.head_commit.message}}' | sed 's/^.*#\([0-9]*\).*/\1/;q')"
+          if [[ $(curl -fs -H 'Accept: application/vnd.github.antiope-preview+json' https://api.github.com/repos/${{github.repository}}/pulls/$pr \
+            | jq '.labels[] | select(.name == "skip upload")') ]]; then
```
sjackman, line 32:
```suggestion
            | jq '.labels[] | select(.name == "ci-skip-upload")') ]]; then
```
Bike shedding here, but how about `ci-skip-upload`. I'm thinking a `ci-` prefix for any labels that have an actual effect on CI, and hyphens instead of spaces to make shell usage easier without having to quote the spaces `gh pr create -lci-skip-upload`

sjackman (changes requested):

.github/workflows/upload-bottles.yml
SeekingMeaning: Shall we rename `ci_error` to `ci-error`?

sjackman: Homebrew/core uses either spaces or hyphens, and I prefer hyphens, so let's go with hyphens. Feel free to rename tags with underscores to hyphens if you like.

sjackman (approved):
Thanks!

Issue #1229, opened 2021-01-04
Emboss bug


dzerbino:
This bug arose when running the testing of new PR, so I do not control the environment. 

- I was running tests on PR #1228 
- What happened (include command output):

```
==> FAILED
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/brewsci/homebrew-bio/Formula/emboss.rb
emboss: Calling depends_on :x11 is deprecated! Use depends_on specific X11 formula(e) instead.
Please report this issue to the brewsci/bio tap (not Homebrew/brew or Homebrew/core), or even better, submit a PR to fix it:
  /usr/local/Homebrew/Library/Taps/brewsci/homebrew-bio/Formula/emboss.rb:26
```

- What you expected to happen

I did not expect Emboss to be an issue as it is not a dependency of that formula. 

- Step-by-step reproduction instructions (by running `brew install` commands)

https://github.com/brewsci/homebrew-bio/pull/1228/checks?check_run_id=1643252910

PR #1230, opened 2021-01-04
emboss: always build without x11

SeekingMeaning:
Closes https://github.com/brewsci/homebrew-bio/issues/1229


sjackman:
Thanks for the fix, Seeker!

SeekingMeaning:
Ideally we’d use libx11 but this was all I could muster 😅

sjackman:
I'd say it's may be better to omit big dependencies, until at least one user asks for it.

Issue #1, opened 2017-10-03
Updates


sjackman:

sjackman:
🎉 New formula arcs in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/arcs
🏡 https://github.com/bcgsc/arcs
📖 https://doi.org/10.1093/bioinformatics/btx675

sjackman:
🎉 New formula minimap2 in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/minimap2
🏡 https://github.com/lh3/minimap2
📖 https://arxiv.org/abs/1708.01492

sjackman:
🎉 New formula miniasm in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/miniasm
🏡 https://github.com/lh3/miniasm
📖 https://doi.org/10.1093/bioinformatics/btw152
🔬 https://github.com/brewsci/homebrew-bio
🐧 http://linuxbrew.sh @linuxbrew #bioinformatics

sjackman:
🎉 New formula canu in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/canu
🏡 https://canu.readthedocs.io/
📖 https://doi.org/10.1101/gr.215087.116

sjackman:
🎉 New formula busco in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/busco
🏡 http://busco.ezlab.org
📖 https://doi.org/10.1093/bioinformatics/btv351

sjackman:
🎉 New formula gepard in Brewsci/bio for Linux and macOS!
🍺 brew install brewsci/bio/gepard
🏡 http://cube.univie.ac.at/gepard
📖 https://doi.org/10.1093/bioinformatics/btm039

sjackman:
🎉 New formula gatk in Brewsci/bio for Linux and MacOS!
🍺 brew install brewsci/bio/gatk
🏠 https://software.broadinstitute.org/gatk/
📖 https://doi.org/10.1101/gr.107524.110

sjackman:
🎉 New formula metaphlan in Brewsci/bio for Linux and MacOS!
🍺 brew install brewsci/bio/metaphlan
🏠 https://huttenhower.sph.harvard.edu/metaphlan
📖 https://doi.org/10.1038/nmeth.2066

sjackman:
[...]

PR #1247, opened 2021-01-19
Create update-tracker workflow

sumanthratna:
automatically posts to https://github.com/brewsci/homebrew-bio/issues/1 when a new formula is merged

(untested)

### DIFF
--- a/.github/workflows/update-tracker.yml
+++ b/.github/workflows/update-tracker.yml
@@ -1,0 +1,40 @@
+name: update-tracker
+
+on:
+  pull_request:
+    branches:
+      - develop
+    types: [closed]
+  workflow_dispatch:
+    inputs:
+      formula_name:
+        description: 'Formula name'
+        required: true
+        default: ''
+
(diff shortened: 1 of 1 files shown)


sjackman:
Neat! There's a Homebrew/linux-dev command `brew announce plink2` that generates the text for these announcements, which I then copy-and-paste into the GitHub issue and into Twitter (https://twitter.com/brew_sci). If it were automated using GitHub Actions, I'd probably prefer to use the `brew announce` command to do that, to avoid duplicating this code. That command could be taught how to post to a GitHub issue. Something like `brew announce --post plink2`. If you want to go down that bunny hole. If not, it's okay to close this PR. There's only a few new formulae per month, so I'm also fine with the manual system.

See https://github.com/Homebrew/homebrew-linux-dev/blob/master/cmd/announce.rb

sumanthratna:
I doubt I'll have the time to get it right so I'll close this for now :)

PR #1278, opened 2021-02-11
faqcs 2.10 (new formula)

lmrodriguezr:

-----


Formula/faqcs.rb
```diff
@@ -0,0 +1,42 @@
+class Faqcs < Formula
+  # cite Lo_2014: "https://doi.org/10.1186/s12859-014-0366-2"
+  desc "Quality Control of Next Generation Sequencing Data"
+  homepage "https://github.com/LANL-Bioinformatics/FaQCs"
+  url "https://github.com/LANL-Bioinformatics/FaQCs/archive/2.10.tar.gz"
+  sha256 "0458e3500adab5257ba11d1db5402adfc52feb936da39c0a26089cbfc1007832"
+  license "GPL-3.0-or-later"
+
+  depends_on "jellyfish"
+  depends_on "r"
```
sjackman, line 10: Is R a hard dependency or an optional dependency? On macOS I prefer to use the R cask rather than the formula, because the cask can use pre-compiled binary R packages, but the formula cannot. The easiest way to do that is to either omit the dependency `r` or add
```ruby
on_linux do
  depends_on "r"
end
```

sjackman:
Thanks again, Luis! 🏆

Formula/faqcs.rb
lmrodriguezr: It requires R for plotting, which is a core feature, but I think it could be used without it. Should we require `r` only `on_linux` better? Thanks for looking over it!

sjackman: That's my preference! Feel free to open a PR.

Issue #1186, opened 2020-10-30
Update gcc dependencies to gcc@9


sjackman:
1. Bump `revision` if present or otherwise add `revision 1`
2. Change `depends_on "gcc"` or `depends_on "gcc@8"` to `depends_on "gcc@9.`

- [x] arcs
- [ ] arks
- [ ] bali-phy
- [ ] bamm
- [ ] biobloomtools
- [ ] bracken
- [ ] canu
- [ ] centrifuge
- [ ] dida
- [ ] discovardenovo
- [ ] ema
- [ ] fastani
- [ ] gappa
- [ ] gfakluge
- [ ] hh-suite
- [ ] idba
- [ ] iqtree
- [ ] iqtree2
- [ ] kmc
- [ ] kraken
- [ ] kraken2
- [ ] megahit
- [ ] metabat
- [ ] mpboot
- [ ] nanopolish
- [ ] ngmerge
- [ ] ntcard
- [ ] ntedit
- [ ] nthits
- [ ] parasail
- [ ] portcullis
- [ ] pplacer
- [ ] r8s
- [ ] raptor
- [ ] raxml
- [ ] seqan@3
- [ ] stacks
- [ ] trinity
- [ ] vcflib
- [ ] viennarna
- [ ] vt
- [ ] wish

PR #1088, opened 2020-07-17
fastani 1.31

SeekingMeaning:
Created with `brew bump-formula-pr`.

PR #1309, opened 2021-03-13
fastani: Bump revision for gcc@9

lmrodriguezr:
- See #1186
- Also: compress test data

-----

### DIFF
--- a/Formula/fastani.rb
+++ b/Formula/fastani.rb
@@ -5,6 +5,7 @@ class Fastani < Formula
   url "https://github.com/ParBLiSS/FastANI/archive/v1.32.tar.gz"
   sha256 "f66b3bb5b272aa3db4deae782e47020fa9aae63eecbbd6fd93df876237d3b2e5"
   license "Apache-2.0"
+  revision 1
   head "https://github.com/ParBLiSS/FastANI.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


lmrodriguezr:
The formula didn't explicitly require `gcc`, so I added the requirement in the `on_macos` block. I suppose it was chain-required by `boost`?

In any case, please take a look in case I can implement a better solution.

Formula/fastani.rb
```diff
@@ -20,6 +21,7 @@ class Fastani < Formula
   uses_from_macos "zlib"
 
   on_macos do
+    depends_on "gcc@9"
```
sjackman, line 24:
```suggestion
```
`fastani` uses `libomp` for OpenMP support in Clang, rather than using GCC, so it does not depend on GCC.
See PR https://github.com/brewsci/homebrew-bio/pull/1088/files
Because of that, this PR is no longer needed.

sjackman (changes requested):
`fastani` does not need to depend on GCC on macOS.

sjackman:
Closing without merging, because `fastani` uses `libomp` rather than `gcc` on macOS, but thanks for this PR all the same for this PR, Luis.

lmrodriguezr:
The `gzip` of the input FastA files reduces the installation to ~1/4, do you think it's worth it? I can prepare a separate commit (unless you think it may cause any issues)

sjackman:
Sounds great! Thanks, Luis.

PR #1318, opened 2021-03-23
apbspdb2pqr: change tag vAPBS-1.5.0

YoshitakaMo:

-----

change the tag to "vAPBS-1.5.0" from "apbs-1.5" because the upstream repository has modified.


YoshitakaMo:
Because pdb2pqr 1.5.0 depends on python2, I couldn't build the formula on LinuxOS.

sjackman:
One option is to add `depends_on :macos` if you want to drop support for Linux. I don't like dropping support for one OS when both were previously supported, but that's a valid option for formulae that have a hard dependency on Python 2.

PR #1319, opened 2021-03-24
Added Formula for BioInterchange.

indiedotkim:
Draft pull request for @sjackman to check. Audit fails right now, but this is to see why I can't end the formula with a `make install` and need a `bin.install` instead.

### DIFF
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -1,0 +1,32 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter. Interchange GFF, GVF, VCF files via JSON."
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "16f4bb6a1b5b5d0cb361ad5ab8b2bd5c49cd79ab6d30fa0b8c2922d99cf97600"
+  license "MIT"
+
+  depends_on "autoconf" => :build
+  depends_on "cmake" => :build
+
+  def install
+    # Note: Cannot parallelize, because:
+    # 1. OpenSSL, curl, and LibDocument are all ExternalProjects
+    # 2. LibDocument also pulls & builds CPython
+    ENV.deparallelize
+    system "cmake", "-DCMAKE_BUILD_TYPE=Release", "-G", "Unix Makefiles"
+    system "make", "python"
+    system "make", "openssl-lib"
+    system "make", "curl-lib"
+    system "make", "libdocument-lib"
+    system "make", "biointerchange"
+    system "strip", "biointerchange"
+    bin.install "biointerchange"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


indiedotkim:
Nope. Fixing those errors first.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter. Interchange GFF, GVF, VCF files via JSON."
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "16f4bb6a1b5b5d0cb361ad5ab8b2bd5c49cd79ab6d30fa0b8c2922d99cf97600"
+  license "MIT"
+
+  depends_on "autoconf" => :build
```
sjackman, line 8: Do you need `automake` as well?

sjackman: Does it use both `autoconf` and `cmake`? That's possible, but also unusual.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on "cmake" => :build
+
+  def install
+    # Note: Cannot parallelize, because:
+    # 1. OpenSSL, curl, and LibDocument are all ExternalProjects
+    # 2. LibDocument also pulls & builds CPython
+    ENV.deparallelize
+    system "cmake", "-DCMAKE_BUILD_TYPE=Release", "-G", "Unix Makefiles"
+    system "make", "python"
```
sjackman, line 17: Use `depends_on "python@3.9"` instead if possible.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
...
+    system "make", "openssl-lib"
+    system "make", "curl-lib"
```
sjackman, line 19: Use `depends_on "curl"` instead if possible.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
...
+    system "make", "libdocument-lib"
+    system "make", "biointerchange"
+    system "strip", "biointerchange"
```
sjackman, line 22: This command should not be necessary. The result of `make biointerchange` should be built without debugging symbols.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
...
+    bin.install "biointerchange"
+  end
+
+  test do
+    assert_match "2.0.5+133", shell_output("#{bin}/biointerchange -v")
+    
+    assert_match "       0", shell_output("#{bin}/biointerchange -v | wc -l")
+    assert_match "      15", shell_output("#{bin}/biointerchange -v 3>&1 1>&2 2>&3 | wc -l")
```
sjackman, line 30: You can remove these two tests.

PR #1320, opened 2021-03-24
biointerchange 2.0.5 (new formula)

indiedotkim:
Second try. Pull request for @sjackman to check why I can't use `make install`, but `bin.install` works for me (locally).

### DIFF
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -11,6 +11,10 @@ class Biointerchange < Formula
 
   uses_from_macos "curl" => :build # linked statically
 
+  on_macos do
+    depends_on "libiconv" => :build # linked statically
+  end
+
   def install
     # NOTE: Cannot parallelize, because of ExternalProject in CMakeLists.txt
     ENV.deparallelize
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -9,6 +9,8 @@ class Biointerchange < Formula
   depends_on "openssl@1.1" => :build # linked statically
   depends_on "python@3.9"
 
+  uses_from_macos "curl" => :build # linked statically
+
   def install
     # NOTE: Cannot parallelize, because of ExternalProject in CMakeLists.txt
     ENV.deparallelize
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -1,0 +1,32 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter: an interchange GFF, GVF, VCF files via JSON"
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "36fcd19b39063fab87fb9f24a0752769ada91e81b82eb354ceb65d5147128141"
+  license "MIT"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+  depends_on "googletest" => :test
+  depends_on "curl"
+  depends_on "gettext"
+  depends_on "openssl@1.1"
+  depends_on "python@3.9"
+  on_macos do
+    depends_on "libiconv"
+  end
+
+  def install
+    # NOTE: Cannot parallelize, because of ExternalProject in CMakeLists.txt
+    ENV.deparallelize
+    system "cmake", *std_cmake_args
+    system "make", "libdocument-lib"
+    system "make", "biointerchange"
+    bin.install "biointerchange"
+  end
+
+  test do
+    assert_match "2.0.5+139\n", shell_output("#{bin}/biointerchange -v")
+  end
+end
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -1,12 +1,10 @@
 class Biointerchange < Formula
-  desc "Genomic linked data converter: an interchange GFF, GVF, VCF files via JSON"
+  desc "Genomic linked-data converter: an interchange GFF, GVF, VCF files via JSON-LD"
   homepage "https://github.com/BioInterchange/BioInterchangeC"
   url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
   sha256 "36fcd19b39063fab87fb9f24a0752769ada91e81b82eb354ceb65d5147128141"
   license "MIT"
 
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
   depends_on "cmake" => :build
   depends_on "googletest" => :test
   depends_on "curl"
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -12,6 +12,7 @@ class Biointerchange < Formula
   uses_from_macos "curl" => :build # linked statically
 
   on_macos do
+    depends_on "gettext" => :build # linked statically
     depends_on "libiconv" => :build # linked statically
   end
 
--- a/Formula/biointerchange.rb
+++ b/Formula/biointerchange.rb
@@ -6,14 +6,8 @@ class Biointerchange < Formula
   license "MIT"
 
   depends_on "cmake" => :build
-  depends_on "googletest" => :test
-  depends_on "curl"
-  depends_on "gettext"
-  depends_on "openssl@1.1"
+  depends_on "openssl@1.1" => :build # linked statically
   depends_on "python@3.9"
-  on_macos do
-    depends_on "libiconv"
-  end
 
   def install
     # NOTE: Cannot parallelize, because of ExternalProject in CMakeLists.txt


indiedotkim:
The Linux build illustrates the problem I'm seeing:

    make: *** No rule to make target 'openssl-lib'.  Stop.

That isn't true. There's definitely such a rule.

The OSX build seems to fail because of gettext missing (or, one of the libraries it installs). I'll have to double check that.

indiedotkim:
The Linux build does not seem to build anything, even though the OS X build works.

Linux "fails" with:

    ==> Running Formulae#formula!(brewsci/bio/biointerchange)
    ==> SKIPPED brewsci/bio/biointerchange
    is not needed on Linux, where iconv.h is provided by glibc

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter: an interchange GFF, GVF, VCF files via JSON"
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "6837b4564ebe996d552de552afe40ac28314a3876c3161c8735f3956c494bb57"
+  license "MIT"
+
+  depends_on "autoconf" => :build
```
sjackman, line 8: Does it also need `automake`?

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on "cmake" => :build
+  depends_on "googletest" => :test
+  depends_on "curl"
+  depends_on "gettext"
```
sjackman, line 12:
```suggestion
  on_macos do
    depends_on "gettext"
  end
```

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on "libiconv"
```
sjackman, line 13:
```suggestion
  uses_from_macos "libiconv"
```

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def install
+    # NOTE: Cannot parallelize, because:
+    # 1. OpenSSL, curl, and LibDocument are all ExternalProjects
+    # 2. LibDocument also pulls & builds CPython
+    ENV.deparallelize
+    system "cmake", "-DCMAKE_BUILD_TYPE=Release", "-G", "Unix Makefiles"
+    system "make", "libdocument-lib"
+    system "make", "biointerchange"
+    system "strip", "biointerchange"
```
sjackman, line 25: This command should not be necessary. The result of `make biointerchange` should be built without debugging symbols thanks to Superenv, which removes `-g` from `gcc -g`.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on "libiconv"
+  depends_on "openssl@1.1"
+  depends_on "python@3.9"
+
+  def install
+    # NOTE: Cannot parallelize, because:
+    # 1. OpenSSL, curl, and LibDocument are all ExternalProjects
+    # 2. LibDocument also pulls & builds CPython
+    ENV.deparallelize
+    system "cmake", "-DCMAKE_BUILD_TYPE=Release", "-G", "Unix Makefiles"
```
sjackman, line 22:
```suggestion
    system "cmake", *std_cmake_args
```

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
...
+    system "make", "biointerchange"
+    system "strip", "biointerchange"
+    bin.install "biointerchange"
+  end
+
+  test do
+    assert_match "2.0.5+139\n", shell_output("#{bin}/biointerchange -v")
+
+    assert_match "1\n", shell_output("#{bin}/biointerchange -v | wc -l | tr -d ' '")
+    assert_match "15\n", shell_output("#{bin}/biointerchange -h 3>&1 1>&2 2>&3 | wc -l | tr -d ' '")
```
sjackman, line 33: You can remove these two tests.

sjackman:
> The Linux build illustrates the problem I'm seeing:
> `make: *** No rule to make target 'openssl-lib'.  Stop.`
> That isn't true. There's definitely such a rule.

An empty or missing `Makefile` would cause this same error.

Formula/biointerchange.rb
```diff
@@ -0,0 +1,35 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter: an interchange GFF, GVF, VCF files via JSON"
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "6837b4564ebe996d552de552afe40ac28314a3876c3161c8735f3956c494bb57"
+  license "MIT"
+
+  depends_on "autoconf" => :build
+  depends_on "cmake" => :build
```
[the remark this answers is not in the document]
indiedotkim: Yes, due to the CPython dependency down the road.

Formula/biointerchange.rb
sjackman: We typically do not run strip for Homebrew. I'd prefer to omit it.

indiedotkim: The default seems to be that Homebrew's build doesn't strip. I have removed the line from the Formula, but it's strange from a C/C++ dev perspective. It's actively leaving junk in a release product.

Formula/biointerchange.rb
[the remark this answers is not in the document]
sjackman: Are you sure? I just built this formula from source myself on macOS 10.15 without `autoconf` installed.

Formula/biointerchange.rb
sjackman: Homebrew's Superenv removes debugging symbols by removing `-g` from `gcc -g`.

indiedotkim: You aren't wrong, but the additional symbols come from external sources and are added when linking.

indiedotkim:
@sjackman Where's brew on the test instances? I use it to look up the latest installed packages to prevent hardcoding of version numbers, but on your Linux instances I get:

    /usr/bin/env: ‘brew’: No such file or directory

That's already within a bash login-shell, so I'm not sure why env can't resolve brew. I can make CMake search where it's on my Debian (`/home/linuxbrew/.linuxbrew/bin`), if that's where your brew is installed too.

sjackman:
```console
$ docker run -it --rm homebrew/ubuntu16.04 which brew
/home/linuxbrew/.linuxbrew/bin/brew
```

Formula/biointerchange.rb
```diff
@@ -0,0 +1,32 @@
+class Biointerchange < Formula
+  desc "Genomic linked data converter: an interchange GFF, GVF, VCF files via JSON"
+  homepage "https://github.com/BioInterchange/BioInterchangeC"
+  url "http://indie.kim/assets/biointerchange/biointerchange-2.0.5.tar.gz"
+  sha256 "36fcd19b39063fab87fb9f24a0752769ada91e81b82eb354ceb65d5147128141"
+  license "MIT"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
```
sjackman, line 9: Are you sure that `autoconf` and `automake` are needed? I built this formula from source myself on macOS 10.15 without `autoconf` installed.

indiedotkim: Good catch.

sjackman:
```console
$ brew linkage biointerchange indiedotkim/develop
System libraries:
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/Python (python@3.9)
Dependencies with no linkage:
  curl
  googletest
  openssl@1.1
```
Are you sure that `curl` and `openssl@1.1` are needed? The library `openssl@1.1` is likely not needed, since there is no run-time linkage to it. macOS provides `/usr/bin/curl`, so it's likely not needed on macOS. You can use `uses_from_macos "curl"` to ensure that it's installed on Linux.

indiedotkim:
No, I am not sure. With the removal of an online license check, BioInterchange does not need those anymore. Having said that, CPython may or may not need them. Either when compiling or when linking statically into the biointerchange executable.

I do not really have time to look into this further, since my next priority is to focus on un-coupling CMakeLists from brew again. The current build relies on brew now (at least on OSX), which is not great due to the lack of brew's versioning. I mentioned that in an email to you.

I do not think that accidentally installing two small packages that possibly are not required anymore is not making a big difference. Last time I checked, the compilation is pulling in over 400MB to produce a <200kB executable. That's a lot more than the previous binary-Formula required *including* the example GFF3, GVF, and VCF files that were part of that.

sjackman:
> No, I am not sure. With the removal of an online license check, BioInterchange does not need those anymore. Having said that, CPython may or may not need them. Either when compiling or when linking statically into the biointerchange executable.

I have removed the unnecessary dependencies.

> I do not think that accidentally installing two small packages that possibly are not required anymore is not making a big difference. Last time I checked, the compilation is pulling in over 400MB to produce a <200kB executable. That's a lot more than the previous binary-Formula required including the example GFF3, GVF, and VCF files that were part of that.

The resulting binary packages are 72 kB on macOS and 104 kB on Linux. See the artifacts tab at https://github.com/brewsci/homebrew-bio/runs/2246596921

indiedotkim:
Surely that brought up a bunch of CMake warnings, because CMakeLists.txt still looks for these dependencies and it will generate empty/incomplete variable assignments.

sjackman:
```
fatal error: curl/curl.h: No such file or directory
compilation terminated.
```
`curl` is needed on Linux.

https://github.com/brewsci/homebrew-bio/pull/1320/checks?check_run_id=2249761263#step:5:1171

indiedotkim:
I appreciate being rigorous with the dependencies, but to me it looks like the build fails due to `license.{c,h}`. This did need curl/openssl. It is only still there, because I haven't had the time to decide which parts of that implementation I still want to keep. I could move it onto a branch and be done with it for now, but again, I know that CPython was tricky with its dependencies. Even with the unit-tests in place, I still would want to look over some test-runs manually to make sure to Python calls still work after removing those dependencies.

I rather work on cleaning up the CMake process, deploy apt-packages, Docker images, and DMGs again, plus improve the documentation and examples, instead of fine-tuning this formula.

If the initial submission to homebrew-bio needs to be perfect in terms of dependencies, than I'd just give it a pass. Especially -- as I have pointed out multiple times -- with the current state of brew that no longer supports versioning. The whole build process is at the mercy of other developers at the moment anyway. So, again, a binary-formula would make a whole lot more sense right now.

sjackman:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
ICONV_LIB
    linked by target "biointerchange-static" in directory /tmp/biointerchange-20210401-5809-1fgdv78
```
https://github.com/brewsci/homebrew-bio/pull/1320/checks?check_run_id=2249761308#step:4:140

sjackman:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
INTL_LIB
```
https://github.com/brewsci/homebrew-bio/pull/1320/checks?check_run_id=2251307257#step:4:199

Depends on `gettext` on macOS.

sjackman:
Merged. Than you for your first contribution to Brewsci/bio, Kim! 🥇

PR #1341, opened 2021-04-13
hh-suite: use clang, add test, add arm64, drop unsupported scripts

milot-mirdita:

-----

For some reason the gcc based binaries are really slow when we tried them. I updated the recipe to correctly use clang with openmp and while doing that generally modernized the recipe.


sjackman:
> For some reason the gcc based binaries are really slow when we tried them.

I'd be curious to know why that is if you ever dig into it.

sjackman:
Thanks again for your contributions to Brewsci/bio, Milot! 🌟

milot-mirdita:
> > For some reason the gcc based binaries are really slow when we tried them.
> 
> I'd be curious to know why that is if you ever dig into it. Only on macOS? Is performance as expected on Linux?

A user told me that their hh-suite on macOS was surprisingly slow. I didn't reproduce the issue myself. It might have also been some completely unrelated problem. However, I know that the homebrew-core recipe for MMseqs2 works well, and decided to backport all those changes to the hh-suite recipe. I can update this issue if I hear back from the user.

PR #1360, opened 2021-06-01
arcs 1.2.2

lcoombe:

-----

### DIFF
--- a/Formula/arcs.rb
+++ b/Formula/arcs.rb
@@ -1,10 +1,10 @@
 class Arcs < Formula
   # cite Yeo_2017: "https://doi.org/10.1093/bioinformatics/btx675"
-  desc "Scaffold genome sequence assemblies using 10x Genomics data"
+  desc "Scaffold genome sequence assemblies using linked or long reads"
   homepage "https://github.com/bcgsc/arcs"
-  url "https://github.com/bcgsc/arcs/releases/download/v1.2.1/arcs-1.2.1.tar.gz"
-  sha256 "c86e2dae359b38bed0a628e60e47a95e496ac9ef0fbda712d9246e7f0332c832"
-  license "GPL-3.0"
+  url "https://github.com/bcgsc/arcs/releases/download/v1.2.2/arcs-1.2.2.tar.gz"
+  sha256 "9c3490eb77be198d28ca55eb10bae617f694f0abee0484d3f599e3354e97450b"
+  license "GPL-3.0-only"
 
   bottle do
     root_url "https://ghcr.io/v2/brewsci/bio"
@@ -20,6 +20,7 @@ class Arcs < Formula
 
   depends_on "boost" => :build
   depends_on "google-sparsehash" => :build
+  depends_on "links-scaffolder"
 
   uses_from_macos "zlib"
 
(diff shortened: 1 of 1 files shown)


Formula/arcs.rb
```diff
@@ -20,6 +20,7 @@ class Arcs < Formula
 
   depends_on "boost" => :build
   depends_on "google-sparsehash" => :build
+  depends_on "links-scaffolder"
```
sjackman, line 23:
```suggestion
```
I'm inclined to omit `links-scaffolder` as a dependency, unless it's required to build ARCS.

sjackman (changes requested):

Formula/arcs.rb
```diff
@@ -40,9 +41,17 @@ def install
       "--prefix=#{prefix}",
       "--with-boost=#{Formula["boost"].opt_include}"
     system "make", "install"
+    libexec_src = Pathname.new("#{libexec}/bin/src")
+    libexec_src.install "src/long-to-linked-pe"
+    libexec_bin = Pathname.new("#{libexec}/bin/Examples")
+    libexec_bin.install "Examples/makeTSVfile.py"
+    libexec_bin.install "Examples/arcs-make"
+    (bin/"arcs-make").write_env_script libexec/"bin/Examples/arcs-make", PYTHONPATH: ENV["PYTHONPATH"]
```
sjackman, line 49:
```suggestion
    libexec.install "src/long-to-linked-pe"
    libexec.install "Examples/makeTSVfile.py"
    libexec.install "Examples/arcs-make"
    (bin/"arcs-make").write_env_script libexec/"arcs-make", PYTHONPATH: ENV["PYTHONPATH"]
```
I'd suggest installing these three files in `libexec`. The directory names `libexec/bin/src` and `libexec/bin/Examples` seem unnecessarily long and clunky.

Formula/arcs.rb
```diff
@@ -40,9 +41,17 @@ def install
...
   end
 
   test do
     assert_match "Usage", shell_output("#{bin}/arcs --help")
+    assert_match "Usage", shell_output("#{bin}/long-to-linked-pe --help 2>&1")
```
sjackman, line 54: I don't see in the formula where `#{bin}/long-to-linked-pe` is installed. Is it installed by `make install`? The formula installs it to `libexec/bin/src` in `libexec_src.install "src/long-to-linked-pe"`, but this test passes, so it must also be installed to `bin`. It ought to be named with an `arcs-` prefix if it's being installed to `bin`, such as `arcs-long-to-linked-pe`.

Formula/arcs.rb
lcoombe: I know it might not be as clean, but they are formatted like that because that's where the `arcs-make` makefile expects them to be (relative to where it is)... We could change for a future release, but that's how it's organized right now.

Formula/arcs.rb
lcoombe: Yes, it is installed by `make install`. I can also make a note of your suggestion for naming for a future release - I wouldn't need it to necessarily be installed to `bin`, but I didn't see how to avoid that since it's installed by `make install`?

Formula/arcs.rb
lcoombe: My reasoning here is that the `arcs-make` Makefile, which most users use as the driver for running ARCS, requires LINKS

sjackman: I'm inclined to omit soft dependencies like this, although I agree that this is a grey-area case.

Formula/arcs.rb
sjackman: Adding an `arcs-` prefix to and installing it to `bin` is easiest (rather than `libexec`).

Formula/arcs.rb
```diff
@@ -40,9 +41,17 @@ def install
       "--prefix=#{prefix}",
       "--with-boost=#{Formula["boost"].opt_include}"
     system "make", "install"
+    libexec_src = Pathname.new("#{libexec}/bin/src")
+    libexec_src.install "src/long-to-linked-pe"
```
sjackman, line 45: Is there both a binary executable `long-to-linked-pe` and a script also named `long-to-linked-pe`?

Formula/arcs.rb
sjackman: Add these scripts to `dist_bin_SCRIPTS` in `Makefile.am`. See for example https://github.com/bcgsc/abyss/blob/da2d0da486443da9b3d1e107fd495e274b709e90/bin/Makefile.am#L8. It'll make this Homebrew formula way easier to maintain if `make install` installs all the required scripts to `bin`.

sjackman: Please add an `arcs-*` prefix to these executables if they're being installed to `bin`.

PR #1362, opened 2021-06-01
tigmint 1.2.3

lcoombe:

-----

### DIFF
--- a/Formula/tigmint.rb
+++ b/Formula/tigmint.rb
@@ -2,8 +2,8 @@ class Tigmint < Formula
   # cite Jackman_2018: "https://doi.org/10.1186/s12859-018-2425-6"
   desc "Correct misassemblies using linked or long reads"
   homepage "https://bcgsc.github.io/tigmint/"
-  url "https://github.com/bcgsc/tigmint/releases/download/1.2.2/tigmint-1.2.2.tar.gz"
-  sha256 "89c03d9b44ffca37f32aee7179b79b74e4f89feba935af74bf6d1b8ff1e17f54"
+  url "https://github.com/bcgsc/tigmint/releases/download/v1.2.3/tigmint-1.2.3.tar.gz"
+  sha256 "6bc72e23de660a0ad3042eeedb9a2b57ffef86c94dadeaaf926c582b458dd297"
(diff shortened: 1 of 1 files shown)


Formula/tigmint.rb
```diff
@@ -17,23 +17,27 @@ class Tigmint < Formula
...
+  depends_on "samtools"
 
   def install
     ENV.prepend_path "PATH", libexec/"bin"
     xy = Language::Python.major_minor_version "python3"
     ENV.prepend_path "PYTHONPATH", libexec/"lib/python#{xy}/site-packages"
     inreplace "bin/tigmint-cut", "/usr/bin/env python3", Formula["python@3.9"].bin/"python3.9"
-    inreplace "bin/tigmint-molecule", "/usr/bin/env python3", Formula["python@3.9"].bin/"python3.9"
-    inreplace "bin/long-to-linked", "/usr/bin/env python3", Formula["python@3.9"].bin/"python3.9"
+    inreplace "bin/tigmint_molecule.py", "/usr/bin/env python3", Formula["python@3.9"].bin/"python3.9"
```
sjackman, line 27: For executables installed in `#{bin}`, I prefer the naming scheme `tigmint-*` with no `.py` extension. Renaming executables can also break users' existing shell scripts that assume the old name. Was there a reason for renaming this executable?

Formula/tigmint.rb
```diff
@@ -17,23 +17,27 @@ class Tigmint < Formula
...
+    inreplace "bin/tigmint_molecule_paf.py", "/usr/bin/env python3", Formula["python@3.9"].bin/"python3.9"
```
sjackman, line 28: I'd suggest `tigmint-molecule-paf` for consistency with other executable names in the package.

Formula/tigmint.rb
lcoombe: Yes, so the reason was that in making the `tigmint_molecule_paf.py` script, which is a faster way to infer molecule extents for `tigmint-long`, to avoid code duplication I imported some functions from `tigmint-molecule`. Unfortunately, though, that naming convention doesn't work within python scripts for imports...

PR #1386, opened 2021-08-06
jmol 14.31.49

sumanthratna:
Created with `brew bump-formula-pr`.

PR #1388, opened 2021-08-09
Fix jmol 14.31.49 SHA-256 checksum

sumanthratna:
https://github.com/brewsci/homebrew-bio/pull/1386

CC @sjackman


sjackman:
Merged! Thanks for the quick fix, Sumanth. 🏅

PR #1437, opened 2021-10-27
gemmi 0.4.9 (new formula)

blake-riley:

-----

### DIFF
--- a/Formula/gemmi.rb
+++ b/Formula/gemmi.rb
@@ -1,0 +1,23 @@
+class Gemmi < Formula
+  desc "Macromolecular crystallography library and utilities"
+  homepage "https://project-gemmi.github.io/"
+  url "https://github.com/project-gemmi/gemmi/archive/refs/tags/v0.4.9.tar.gz"
+  sha256 "1e5a4e4351e5860d70c3aa4ebc762b3f098160dad936e1775bd1d165575516bf"
+  license "MPL-2.0"
+  head "https://github.com/project-gemmi/gemmi.git", branch: "master"
+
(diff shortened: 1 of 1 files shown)


sjackman:
```
==> brew linkage --test brewsci/bio/gemmi
==> FAILED
Unwanted system libraries:
  /lib/x86_64-linux-gnu/libz.so.1
```
https://github.com/brewsci/homebrew-bio/runs/4076248655?check_suite_focus=true#step:5:102

This error means that this formula ought to depend on `zlib`.
Because `zlib` is provided by macOS, you can add `uses_from_macos "zlib"`. You'll see examples of this style of dependency in other formulae.

Formula/gemmi.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", "-S", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/gemmi", "--help"
```
sjackman, line 17: You can use `assert_match` to ensure that `gemmi --help` has run correctly (didn't fail). You'll see examples of using `assert_match` and `shell_output` in other formulae.

sjackman (changes requested):

Formula/gemmi.rb
blake-riley: I'm now using assert_matches. Thanks for the pointer.

sjackman (approved):

sjackman:
Thank you for your first contribution to Brewsci/bio, Blake! 🥇

Issue #1703, opened 2024-06-21
PRs fail tests because shellcheck doesn't like github workflows


bredelings:
Running `brew style homebrew-bio` find problems with the scripts `build-bottles.yml` and `upload-bottles.yml`.  This prevents any PRs from passing tests.

```
$ brew style homebrew-bio/
Inspecting 337 files
.................................................................................................................................................................................................................................................................................................................................................

337 files inspected, no offenses detected
homebrew-bio/.github/workflows/build-bottles.yml:23:9: shellcheck reported issue in this script: SC2086:info:1:42: Double quote to prevent globbing and word splitting [shellcheck]
   |
23 |         run: echo "/home/linuxbrew/.linuxbrew/bin" >> $GITHUB_PATH
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:21:9: shellcheck reported issue in this script: SC2086:info:1:42: Double quote to prevent globbing and word splitting [shellcheck]
   |
21 |         run: |
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:27:9: shellcheck reported issue in this script: SC2046:warning:6:21: Quote this to prevent word splitting [shellcheck]
   |
27 |         run: |
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:27:9: shellcheck reported issue in this script: SC2035:info:8:13: Use ./*glob* or -- *glob* so names with dashes won't become options [shellcheck]
   |
27 |         run: |
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:27:9: shellcheck reported issue in this script: SC2086:info:12:137: Double quote to prevent globbing and word splitting [shellcheck]
   |
27 |         run: |
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:27:9: shellcheck reported issue in this script: SC2086:info:18:149: Double quote to prevent globbing and word splitting [shellcheck]
   |
27 |         run: |
   |         ^~~~
homebrew-bio/.github/workflows/upload-bottles.yml:27:88: "github.event.head_commit.message" is potentially untrusted. avoid using it directly in inline scripts. instead, pass it through an environment variable. see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions for more details [expression]
```

The error location is clearly wrong.  I was able to figure out how to fix 5/7 of the problems, but not all of them.

YoshitakaMo:
[...]

PR #1704, opened 2024-06-21
Fix shellcheck failures

bredelings:
This is an attempt to fix the shellcheck failures in #1703

There is one "style" issue left that I'm not sure can be fixed.


bredelings:
This still fails because of the `sed` expression that tries to extract the PR number from the commit message.

YoshitakaMo:
Thank you, @bredelings. I have merged this PR on behalf of the administrator. However, I have not yet found a fundamental solution.

PR #2025, opened 2025-05-23
promod3: Bump up openstructure to v2.10.0

eunos-1128:
- [x] Have you built your formula locally with `HOMEBREW_NO_INSTALL_FROM_API=1 brew install --build-from-source --verbose --keep-tmp ./Formula/<FORMULA>.rb`, where `FORMULA` is the name of the formula you're submitting?
- [x] Does your formula have no offenses with `brew style /path/to/formula.rb`?
- [x] Does your formula pass `brew audit --formula brewsci/bio/<FORMULA> --online --git --skip-style`?
- [x] Does your formula pass `brew linkage --cached --test --strict brewsci/bio/<FORMULA>` after manual installation?

-----

### DIFF
--- a/Formula/promod3.rb
+++ b/Formula/promod3.rb
@@ -1,0 +1,64 @@
+class Promod3 < Formula
+  # cite Studer_2021: "https://doi.org/10.1371/journal.pcbi.1008667"
+  desc "Versatile Homology Modelling Toolbox"
+  homepage "https://openstructure.org/promod3"
+  url "https://git.scicore.unibas.ch/schwede/ProMod3/-/archive/3.4.2/ProMod3-3.4.2.tar.gz"
+  sha256 "8103bcb344489eb0fa0567ad8c9a8a9b42d3dbbb8d46c82587e6a58eab45eefd"
+  license "Apache-2.0"
+
+  depends_on "cmake" => :build
+  depends_on "eigen" => :build
+  depends_on "boost"
+  depends_on "boost-python3"
+  depends_on "brewsci/bio/openmm@7"
+  depends_on "brewsci/bio/openstructure"
+  depends_on "python@3.13"
+
+  def install
+    if OS.mac?
+      ENV.prepend "LDFLAGS", "-undefined dynamic_lookup -Wl,-export_dynamic"
+    elsif OS.linux?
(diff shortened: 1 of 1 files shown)


eunos-1128:
It is ready to be merhed.

YoshitakaMo:
Sorry, I forgot to mention that you have to add `revision 1` after the `license` line in this formula. Without this, embedding the updated bottle into the formula will be rejected because the same bottle name already exists.  
Also, please add `revision 1` (or `2`?) for `qmean` and create a new PR.

eunos-1128:
@YoshitakaMo 

There seem to be two options, `rebuild` and `revision`. Is `revision` appropriate in this case? I feel that `rebuild` would be more appropriate since it is a change of dependency.

YoshitakaMo:
Now we are encountering an error when uploading the qmean bottle. See https://github.com/brewsci/homebrew-bio/actions/runs/15227450548. At this time, incrementing the value of `rebuild` might be appropriate according to Homebrew's rules. If

```bash
bottle do
  rebuild 1
end
```

is added to the formula and it allows the Homebrew CI to pass, that would be desirable. However, I am not sure if it will resolve the error mentioned above.

Additionally, we have taken the approach of deleting the Formula for qmean and promod3 once. In this case, it is unclear which method would be more appropriate.
When I encountered a similar situation before, I resolved the conflict with the previously uploaded bottle by adding `revision 1`.

Please try incrementing the `rebuild` first, and if the bottle can be uploaded correctly, that would be fine. However, if that does not work, please try using `revision` instead.

eunos-1128:
Would it be possible to merge them?

YoshitakaMo:
According to the [actions log](https://github.com/brewsci/homebrew-bio/actions/runs/15239202834), the insertion of this block resulted in no effect:

> ```
> bottle do
>   rebuild 1
> end
> ```

However, the second `qmeans` [PR](https://github.com/brewsci/homebrew-bio/pull/2028) was successfully merged. The internal rebuild count was set to `1` in this PR because [a preceding PR](https://github.com/brewsci/homebrew-bio/actions/runs/15220252054) was merged in advance. Could you please resubmit the same formula in another PR?

PR #2190, opened 2026-06-21
salmon 2.1.2

BenjaminDEMAILLE:
Update salmon to the 2.x line.

salmon 2.0 is a from-scratch **Rust rewrite** of salmon, shipped as **prebuilt per-platform binaries** via cargo-dist; the final C++ release (1.10.x) now lives on the upstream [`cpp`](https://github.com/COMBINE-lab/salmon/tree/cpp) branch. The workflow and output formats are unchanged (`salmon index` -> `salmon quant` -> `quant.sf`), but the build and license changed.

### Changes
- Install the **prebuilt artifact for the host platform** (macOS/Linux x arm64/x86_64) from the GitHub release, instead of building from source: the old C++ build deps (autoconf, automake, boost, cmake, tbb, xz, zstd, and the `uses_from_macos` set) are gone, and there is no compile step.
- License updated to `BSD-3-Clause` (was `GPL-3.0-or-later`), matching the rewrite.
- Added a `livecheck` block (`github_latest`) so future releases are picked up by `brew livecheck` / autobump.
- Dropped the stale C++ `bottle` block.

### Validation (local, arm64 macOS)
- `brew style Formula/salmon.rb` — no offenses
- `brew audit --strict --online --formula brewsci/bio/salmon` — clean
- `brew install brewsci/bio/salmon` — pours the prebuilt binary (instant)
- `brew test brewsci/bio/salmon` — `salmon --version` reports `salmon 2.1.1`, and `salmon index` runs end to end

### Note
If you'd rather build from source for bottling, I'm happy to switch to a `depends_on "rust" => :build` + `cargo install --path crates/salmon-cli` formula instead (also validated locally, ~3 min build). I went with the prebuilt binaries since salmon already publishes them per platform via cargo-dist. Per-artifact sha256s are in the release's `sha256.sum`.

### DIFF
--- a/Formula/salmon.rb
+++ b/Formula/salmon.rb
@@ -13,24 +13,30 @@ class Salmon < Formula
   # salmon 2.0 is a from-scratch Rust rewrite shipped as a single binary via
   # cargo-dist (the final C++ release, 1.10.x, lives on the upstream `cpp`
   # branch). Use the prebuilt per-platform artifacts directly.
+  #
+  # The artifact filenames (salmon-cli-<target-triple>.tar.xz) carry no version,
+  # and the `x86_64` triple makes Homebrew mis-scan the version as
+  # "64-unknown-linux-gnu". The `#/salmon.tar.xz` fragment renames the download
+  # so the scanner ignores the triple and picks up "2.1.1" from the URL path on
+  # every platform, keeping detection consistent without a redundant `version`.
   on_macos do
     on_arm do
-      url "https://github.com/COMBINE-lab/salmon/releases/download/v2.1.1/salmon-cli-aarch64-apple-darwin.tar.xz"
+      url "https://github.com/COMBINE-lab/salmon/releases/download/v2.1.1/salmon-cli-aarch64-apple-darwin.tar.xz#/salmon.tar.xz"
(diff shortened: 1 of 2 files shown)


YoshitakaMo:
`brew test` failed: https://github.com/brewsci/homebrew-bio/actions/runs/27899034874/job/82629299134#step:8:56
Please check it.

BenjaminDEMAILLE:
> `brew test` failed: https://github.com/brewsci/homebrew-bio/actions/runs/27899034874/job/82629299134#step:8:56 Please check it.

solved @YoshitakaMo

BenjaminDEMAILLE:
Heads up: the current red checks here are **not** from this PR's changes. They're the tap-wide `brew style brewsci/bio` failure caused by Homebrew newly enforcing the `Homebrew/FormulaPathMethods` cop (200 offenses across ~50 untouched formulae). Tracked in #2192.

The salmon + trinity changes in this branch pass on their own (salmon resolves to `2.1.1` on all platforms; trinity gains `DB_File` on Linux). Once the tap-wide style issue is resolved, I'll rebase. Earlier salmon `salmon.rb#L42` / `64-unknown-linux-gnu` annotations are stale from the first run, before the fix.

BenjaminDEMAILLE:
@YoshitakaMo done for this one

PR #2200, opened 2026-06-26
Add autobump workflow

BenjaminDEMAILLE:
Adds a scheduled GitHub Actions workflow (`.github/workflows/autobump.yml`) that automatically opens version-bump PRs for outdated formulae.

### How it works
- Runs weekly (Mondays 05:00 UTC) and on manual `workflow_dispatch`.
- Sets up Homebrew, taps this repo, and runs `brew bump --tap brewsci/bio --formulae --open-pr`.
- `brew bump` uses each formula's existing `livecheck` block to detect new upstream releases, then opens a PR per outdated formula. Formulae that already have an open bump PR are skipped.

### Setup required (otherwise inert)
- Add a repository secret `HOMEBREW_GITHUB_API_TOKEN`: a PAT (classic `public_repo`/`repo`, or fine-grained with Contents + Pull requests write) for the account that should author the PRs. A PAT (rather than the default `GITHUB_TOKEN`) is used so the bump PRs trigger the existing `Build bottles` CI.

### Notes
- The workflow does nothing until the secret is set, so merging is safe.
- The first run may open several PRs at once (one per currently-outdated formula); adjust the cron cadence or scope to specific formulae if that is too much.

### DIFF
--- a/.github/workflows/autobump.yml
+++ b/.github/workflows/autobump.yml
@@ -1,0 +1,48 @@
+name: Autobump
+
+# Opens version-bump pull requests for outdated formulae, using each formula's
+# `livecheck` block to detect new upstream releases.
+#
+# Requires a repository secret named `HOMEBREW_GITHUB_API_TOKEN`: a Personal
+# Access Token (classic `public_repo`/`repo` scope, or a fine-grained token with
+# Contents + Pull requests write) belonging to the account that should open the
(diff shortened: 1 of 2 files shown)


BenjaminDEMAILLE:
@YoshitakaMo

YoshitakaMo:
Thank you for your great PRs!

PR #2202, opened 2026-06-26
vep 116.0 (new formula)

BenjaminDEMAILLE:
New formula for the [Ensembl Variant Effect Predictor](https://www.ensembl.org/info/docs/tools/vep/index.html) (VEP), `release/116.0`.

The point of this formula is that `brew install vep` gives a working `vep` with **all Perl/API dependencies bundled**, so users don't have to run `INSTALL.pl` and fight its dependency installation.

### What's bundled
- The **Ensembl Perl API** (`ensembl`, `ensembl-variation`, `ensembl-funcgen`, `ensembl-io`), pinned to the `release/116` branch HEAD commits.
- **BioPerl** (the version frozen by the API, staged as pure Perl).
- CPAN deps via `cpanm` into `libexec/vendor`, including the two XS bindings:
  - **Bio::DB::HTS** built against Homebrew `htslib`.
  - **DBD::mysql** (database mode) built against `mysql-client` (+ `zstd`/`openssl@3`).
- `vep`, `filter_vep`, `haplo`, `variant_recoder` and `vep_install` wrapped with the right `PERL5LIB`/`PATH`.

Cache and FASTA data are not bundled (tens of GB); the caveats show how to fetch them with the bundled `vep_install`.

### Notes
- Newer binding versions are used than `INSTALL.pl`'s pins (Bio::DB::HTS 3.01, DBD::mysql 5.013), because the pinned ones (htslib 1.9 / Bio::DB::HTS 2.11, DBD::mysql <= 4.050) do not build against Homebrew's current `htslib`/`mysql-client`.

### Verification (macOS arm64)
- `brew install --build-from-source vep` OK
- `brew test vep` OK (`vep --help`, `filter_vep --help`)
- End-to-end smoke: `vep --offline --cache` loads the full pipeline (Ensembl API + Bio::DB::HTS) and stops only at the expected "cache not found".
- `brew audit --new --strict --online vep` OK
- `brew style vep` OK

### DIFF
--- a/Formula/vep.rb
+++ b/Formula/vep.rb
@@ -1,0 +1,114 @@
+class Vep < Formula
+  # cite McLaren_2016: "https://doi.org/10.1186/s13059-016-0974-4"
+  desc "Ensembl Variant Effect Predictor (VEP)"
+  homepage "https://www.ensembl.org/info/docs/tools/vep/index.html"
+  url "https://github.com/Ensembl/ensembl-vep/archive/refs/tags/release/116.0.tar.gz"
+  sha256 "618a4b6d37efbe0968d7ad1115bf6b712f8537c4697659be6c41580708eb5167"
+  license "Apache-2.0"
+
+  livecheck do
+    url "https://github.com/Ensembl/ensembl-vep.git"
+    strategy :git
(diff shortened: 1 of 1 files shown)


BenjaminDEMAILLE:
@YoshitakaMo

PR #2203, opened 2026-06-27
raptor 3.0.1

BenjaminDEMAILLE:
Bumps `raptor` 2.0.0 → 3.0.1 (latest upstream release).

Supersedes #1619, which targeted 3.0.0 and had gone stale. Changes vs. that PR:

- Targets the newer **3.0.1** tag.
- The macOS CI runners are now all arm64, but the formula forced the x86-only `-mavx2` flag, which cannot compile on Apple Silicon. The flag is now applied on Intel hardware only, so arm64 builds without it.
- Requires `gcc>=10` (`gcc@12`) and adds the Mehringer_2023 citation, as in the original PR.

Bottle block left stale on purpose per the tap convention (CI rebuilds and the post-merge automation refreshes bottles).

### DIFF
--- a/.github/workflows/build-bottles.yml
+++ b/.github/workflows/build-bottles.yml
@@ -31,12 +31,13 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@main
 
-      - name: Trust external taps
-        run: |
-          brew tap aws/tap
-          brew tap azure/bicep
-          brew trust aws/tap
-          brew trust azure/bicep
+      - name: Remove pre-installed external taps
+        # The GitHub runner images ship the aws/tap and azure/bicep taps, which
+        # are irrelevant to this tap. They were previously trusted to satisfy
+        # brew's tap-trust requirement, but `brew trust` now fails on Linux with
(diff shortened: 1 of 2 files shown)


YoshitakaMo:
No bottles were built on macOS. Please check for them.

BenjaminDEMAILLE:
Pushed a fix for the missing macOS bottle. Three issues were stacked:

1. gcc@12's bundled `<arm_acle.h>` is broken on arm64 macOS (Tahoe): `invalid conversion from 'uint64_t*'`.
2. The newest brewed gcc (16) mis-emits duplicate `seqan3::add_enum_bitwise_operators` template symbols that Apple's linker rejects.
3. The brewed yaml-cpp (clang/libc++) is ABI-incompatible with this gcc/libstdc++ build, so chopper/sharg fail to link YAML symbols.

The formula now compiles with gcc@14, builds a matching yaml-cpp from source on macOS, and declares the `gcc` runtime so `brew linkage --test` passes. Linux is unchanged (gcc@12 + brewed yaml-cpp). Verified locally on arm64 Tahoe: builds, passes the test block, linkage clean.

PR #2213, opened 2026-06-27
multiqc 1.35 (new formula)

BenjaminDEMAILLE:
New formula for [MultiQC](https://multiqc.info/) v1.35, which aggregates bioinformatics QC reports across many samples and tools.

Python `Language::Python::Virtualenv` formula on `python@3.13` with 55 bundled resources. Notes for CI:

- **`rust` (build):** required to compile `pydantic-core`, `polars`, `rpds-py`, `tiktoken`.
- **`jpeg-turbo` + `freetype` + `zlib`:** required to build `pillow` from source.
- **Wheel-only resources:** `kaleido` (no sdist) and the two `polars-runtime-*` crates (pinned to a nightly toolchain, can't build on stable Rust) are installed from upstream wheels; everything else builds from source. Wheels cover macOS-arm64 + Linux (arm64/x86_64), matching the arm64-macOS + ubuntu CI matrix.

Verified locally on arm64 macOS: builds in ~3m42s and `multiqc --version` reports `1.35`; `brew test`/`brew style` pass.

### DIFF
--- a/.github/workflows/build-bottles.yml
+++ b/.github/workflows/build-bottles.yml
@@ -31,12 +31,13 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@main
 
-      - name: Trust external taps
-        run: |
-          brew tap aws/tap
-          brew tap azure/bicep
-          brew trust aws/tap
-          brew trust azure/bicep
+      - name: Remove pre-installed external taps
+        # The GitHub runner images ship the aws/tap and azure/bicep taps, which
+        # are irrelevant to this tap. They were previously trusted to satisfy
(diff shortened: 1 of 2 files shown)


BenjaminDEMAILLE:
Pushed a fix for the ubuntu-latest failure. `brew linkage --test` was rejecting system libraries (nss, nspr, expat, sqlite) pulled in by kaleido 0.2.1's bundled headless-chromium; declaring the brewed copies cannot satisfy a prebuilt binary's hardcoded `/lib` paths. kaleido is now restricted to macOS (where its wheels link only allowed frameworks). On Linux, MultiQC still produces its interactive HTML reports; only static PNG/PDF/SVG export is omitted.

PR #2225, opened 2026-07-06
Autobump: run hourly, tolerate per-formula failures, add direct GitHub-tag fallback

BenjaminDEMAILLE:
The weekly Autobump runs have been failing (exit 1) because a single formula's
error aborts the whole job — e.g. `webin-cli` (livecheck reports a bogus `9.0.3`
→ 404) and `voronota` (`brew audit` failure). Valid bump PRs still get opened,
but the run is marked red every week.

This makes the workflow more reactive and resilient:

1. **Hourly schedule.** GitHub Actions cannot subscribe to external upstream
   repos' release events, so the cron is moved from weekly to hourly to
   approximate "bump as soon as a new version is published". `brew bump`
   de-duplicates, so it never re-opens an existing PR.

2. **Tolerate per-formula failures.** The `brew bump --open-pr` step now runs
   with `continue-on-error: true` and writes a step summary. Dead upstreams,
   misdetected versions, or audit failures no longer fail the scheduled run;
   the valid bump PRs are still opened.

3. **Direct GitHub-tag fallback.** A new step
   (`.github/scripts/autobump-git-direct.sh`) handles formulae whose `livecheck`
   returns "Unable to get versions" by reading the upstream GitHub tags directly
   and opening a bump PR when a newer stable tag exists. Non-GitHub sources are
   skipped; `brew bump-formula-pr` validates the tag on download.

The required `HOMEBREW_GITHUB_API_TOKEN` secret is unchanged.

### DIFF
--- a/.github/scripts/autobump-git-direct.sh
+++ b/.github/scripts/autobump-git-direct.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+#
+# Fallback autobump for formulae that livecheck cannot resolve.
+#
+# `brew bump` relies on each formula's `livecheck` block. For formulae whose
+# upstream cannot be checked that way (livecheck reports "Unable to get
+# versions"), this script reads the upstream GitHub tags directly and opens a
+# version-bump PR when a newer *stable* tag exists.
+#
+# It is best-effort: every step tolerates failures so one bad formula never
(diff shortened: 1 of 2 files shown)


BenjaminDEMAILLE:
@YoshitakaMo

PR #2249, opened 2026-07-07
libcifpp 10.0.4

BenjaminDEMAILLE:
Bump `libcifpp` to 10.0.4. Detected via `brew livecheck`.


YoshitakaMo:
I will no longer maintain libcifpp because dssp already includes it.

PR #2289, opened 2026-07-07
uniqtag 1.0.1

BenjaminDEMAILLE:
Bump `uniqtag` to 1.0.1. Detected via `brew livecheck`.


YoshitakaMo:
The build-bottles process was incomplete.
https://github.com/brewsci/homebrew-bio/actions/runs/28976504654?pr=2282
We need to fix the workflow file to check whether all the bottles were correctly built before or during the "merge" step.

PR #2313, opened 2026-07-08
autobump: fix shellcheck SC2086 breaking tap-syntax on every PR

BenjaminDEMAILLE:
#2225 was a workflow-only PR so it skipped build-bottles CI, and a shellcheck issue slipped in: the `${FORMULAE}` expansion at autobump.yml:74 is unquoted (SC2086). This makes the whole-tap `brew style brewsci/bio` (run in every PR's tap-syntax step) fail, so **all open PRs currently fail CI after ~1 min**. Fix: read the dispatch input into an array and quote it.


BenjaminDEMAILLE:
@YoshitakaMo

PR #2325, opened 2026-07-09
gargs 0.3.9 (new formula)

BenjaminDEMAILLE:
New formula for [gargs](https://github.com/brentp/gargs) by Brent Pedersen.

A better(?) `xargs` written in Go: run commands in parallel with convenient
per-argument templating (`{}`, `{0}`, `{1}`, ...). Apache-2.0.

Upstream predates Go modules, so the formula initialises a module at build
time to resolve dependencies.

### Checklist
- [x] `brew install --build-from-source gargs`
- [x] `brew test gargs`
- [x] `brew audit --new --strict gargs`

### DIFF
--- a/Formula/gargs.rb
+++ b/Formula/gargs.rb
@@ -1,0 +1,22 @@
+class Gargs < Formula
+  desc "Better(?) xargs in Go: run commands in parallel with templating"
+  homepage "https://github.com/brentp/gargs"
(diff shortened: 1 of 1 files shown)


YoshitakaMo:
This tap only accepts software related to biology. gargs does not seem to be within its scope.

